PR #20, opened 2025-11-04
feat: add create-account bottom-sheet component

liaugust:


packages/app/src/@transaction/components/create-transaction-bottom-sheet/create-transaction-bottom-sheet.tsx
```diff
@@ -0,0 +1,59 @@
...
+import type { Ref } from 'react';
+
+interface Props {
+    readonly ref: Ref<BottomSheetModal>;
+}
+
+export const CreateTransactionBottomSheet = ({ ref }: Props) => {
+    const { t } = useLingui();
+
+    const transactionTypes = [
```
vitalyiegorov: Let's avoid such approach and instead use JSX, it a lot easier for long term support

vitalyiegorov (changes requested):

PR #46, opened 2025-11-17
fix: update bottom-sheet

liaugust:


packages/app/src/@generic/components/bottom-sheet/bottom-sheet.tsx
```diff
@@ -23,22 +22,40 @@ const Modal = styled(BottomSheetModal, {
...
-        handleIndicatorClassName="bg-primary"
-        ref={ref}
-    >
-        <Content className={cn('bg-primary-reverse pt-4 px-6 pb-6', contentClassName)}>
-            <SafeAreaView edges={edges}>{children}</SafeAreaView>
-        </Content>
-    </Modal>
-);
+export const BottomSheet = ({
+    ref,
```
vitalyiegorov: When we have a long list of destructured params do it as the first operation inside render function to make components definition look beautiful

vitalyiegorov (changes requested):

PR #51, opened 2025-11-17
feat: add language setting

liaugust:


packages/app/src/@generic/interface/language.interface.ts
```diff
@@ -0,0 +1,8 @@
+import { LanguageEnum } from '@budgie/contracts';
+import { MessageDescriptor } from '@lingui/core';
+
+export interface LanguageInterface {
```
vitalyiegorov: Let's extract all i18 related code to i18n module

vitalyiegorov (approved):

PR #58, opened 2025-11-20
feat: add archived accounts screen

liaugust:


packages/app/src/@generic/components/button/button.tsx
```diff
@@ -19,6 +19,7 @@ interface Props extends ComponentProps<typeof HapticPressable> {
     readonly leftIcon?: IconName;
     readonly rightIcon?: IconName;
     readonly size?: ButtonSizeType;
+    readonly textClassName?: string;
```
vitalyiegorov: Can we provide variants?

packages/app/src/app/(tabs)/settings.tsx
```diff
@@ -20,6 +20,7 @@ export default function SettingsPage() {
     const { t } = useLingui();
 
     const navigateToCategories = () => void router.push('/(main)/categories');
+    const navigateToArchived = () => void router.push('/(main)/archived');
```
vitalyiegorov: Can we name all on* handlers with prefix 'handle*' so not as action but event descriptor
handleArxhievedAccountPress

vitalyiegorov (changes requested):
Agree with copilot

vitalyiegorov (approved):

PR #64, opened 2025-12-01
feat: income transaction creation

liaugust:


package.json
```diff
@@ -63,5 +63,8 @@
     "packageManager": "yarn@4.10.3",
     "engines": {
         "node": ">=22.0.0"
+    },
+    "dependencies": {
+        "react-native-ui-datepicker": "^3.1.2"
```
vitalyiegorov: move `app` package?

packages/app/src/account/service/account-balance-incremental.service.ts
```diff
@@ -0,0 +1,81 @@
...
+        const accounts = await accountRepository.getAllActiveAccounts();
+        if (isEmptyArray(accounts)) {
+            return;
+        }
+
+        const accountIds = accounts.map(({ id }) => id);
+
+        const [balanceSnapshots, newEntries] = await Promise.all([
+            accountBalanceRepository.getAccountBalanceSnapshots(accountIds),
+            accountBalanceRepository.getNewTransactionEntries(accountIds)
```
vitalyiegorov: if you already have an optimized SQL queries to calculate account balance, we should use them, so the logic for this dalayed recalculation is to recalculate ALL ENTRIES, not following the same logic of snapshot + entries after date, but instead all of them, to create a correct snapshot and "self-heal" if needed. 

I would also provide this method as a repository method and add this under the settings to force "full recalculation". 

So this class becomes only a "scheduled task" to run certain low level SQL repo code.

packages/app/src/account/service/account.service.ts
```diff
@@ -1,31 +1,113 @@
...
-            const account = await accountRepository.updateById(id, input, tx);
+        return db.transaction(async tx => {
+            const [{ balance: currentBalanceMicro }] = await accountBalanceRepository.getAccountBalance(id);
+
+            const account = await accountRepository.updateById(
+                id,
+                {
+                    ...input,
+                    ...(isNumber(input.currentBalance) && {
+                        currentBalance: convertToMicroUnits(input.currentBalance)
```
vitalyiegorov: same here, we do not have `currentBalance` this is a calculated field.

packages/app/src/account/service/account.service.ts
```diff
@@ -1,31 +1,113 @@
...
-                await accountBalanceRepository.updateByAccountId(account.id, { amount: input.currentBalance }, tx);
+            if (isNumber(input.currentBalance) && convertToMicroUnits(input.currentBalance) !== currentBalanceMicro) {
+                await this.adjustBalanceTo(account.id, account.instrumentId, input.currentBalance, tx);
             }
 
             return account;
         });
     }
+
+    private async adjustBalanceTo(accountId: number, instrumentId: number, targetBalance: number, tx: Transaction): Promise<void> {
```
vitalyiegorov: Lets discuss why do we need this?

packages/app/src/transaction/service/transaction.service.ts
```diff
@@ -0,0 +1,49 @@
+import { TransactionCreateEntityInterface, TransactionEntityInterface } from '@budgie/contracts';
+
+import { isNotEmptyArray } from '@rnw-community/shared';
+
+import { db, transactionEntryRepository, transactionRepository, transactionTagsRepository } from '../../@generic/drizzle/db/db';
+import { convertToMicroUnits } from '../../@generic/utils/convert-to-micro-units.util';
+
+class TransactionService {
+    async createInternal(input: TransactionCreateEntityInterface): Promise<TransactionEntityInterface> {
```
vitalyiegorov: Why this is called `createInterval`?

packages/app/src/transaction/service/transaction.service.ts
```diff
@@ -0,0 +1,49 @@
...
+                    operatedAt: input.operatedAt,
+                    toAccountId: input.toAccountId,
+                    exchangeRate: input.exchangeRate,
+                    fromAccountId: input.fromAccountId,
+                    amount: convertToMicroUnits(input.amount)
+                },
+                tx
+            );
+
+            await Promise.all(
```
vitalyiegorov: Do we need have an input validation that `entries` are not empty?

packages/contracts/src/account-balance/repository/account-balance.repository.ts
```diff
@@ -1,32 +1,170 @@
...
         return accountBalance;
     }
 
-    async updateByAccountId(accountId: number, input: AccountBalanceUpdateEntityInterface, tx?: TX): Promise<AccountBalanceEntityInterface> {
-        const [accountBalance] = await (tx ?? this.db)
-            .update(AccountBalanceEntityTable)
-            .set(input)
-            .where(eq(AccountBalanceEntityTable.accountId, accountId))
-            .returning();
+    async getAccountBalanceSnapshots(accountIds: number[]): Promise<AccountBalanceEntityInterface[]> {
```
vitalyiegorov: this is `AccountBalanceRepository` lets remove this words from the method names, also `snapshot` is a bit confusing, this are just records of this table, lets name it as in other repos, `findAllByAccountId` or similar

packages/contracts/src/account-balance/repository/account-balance.repository.ts
```diff
@@ -1,32 +1,170 @@
...
+                    ),
+                    0
+                )
+            `
+            })
+            .from(AccountEntityTable)
+            .where(and(eq(AccountEntityTable.includeInNetWorth, true), isNull(AccountEntityTable.deletedAt)));
+    }
+
+    private getAccountBalanceWithTransactions() {
```
vitalyiegorov: we need to use word `Sql` -> `getAccountBalanceWithTransactionsSql` to differentiate this methods from the rest

packages/contracts/src/account-balance/repository/account-balance.repository.ts
```diff
@@ -1,32 +1,170 @@
...
+        return sql`
+            COALESCE(${this.getLatestAccountBalance()}, 0)
```
vitalyiegorov: also i guess we can inline the `COALESCE` into those SQL methods?

packages/app/src/transaction/service/transaction.service.ts
liaugust: This is do differentiate transactions that will come from external sources like "mono", "pumb", etc

packages/app/src/transaction/service/transaction.service.ts
liaugust: We do have it on the zod schema level

vitalyiegorov (approved):

PR #67, opened 2025-12-14
feat: add create expense transaction

liaugust:


liaugust:
@copilot Fill empty translations for lingui

vitalyiegorov (approved):

PR #169, opened 2026-01-02
fix: monobank forward sync, optimize transaction query

vitalyiegorov:


vitalyiegorov:
@copilot Please provide missing translations

PR #233, opened 2026-01-09
fix(app): fix tag creation crash

### LINKED ISSUE #224: fix(app): tag creation crashes app

vitalyiegorov:
## Summary
Fixes #224 - Tag creation crashes the app

## Root Cause
The `TagFormBottomSheet` component was using a different pattern for handling form submission compared to the working `CategoryFormBottomSheet`:

**Before (broken):**
```typescript
const onSubmit = () => void handleSubmit(createTag)();
```

**After (fixed):**
```typescript
const onSubmit = handleSubmit(async (values: TagCreateEntityInterface) => {
    // ... handle creation
});
```

## Changes
- Aligned `TagFormBottomSheet` to use the same `handleSubmit` pattern as `CategoryFormBottomSheet`
- Inlined the async handler directly in the `handleSubmit` call

## Testing
- Tag creation from the tags selector now works correctly
- Verified category creation still works (was already using correct pattern)


vitalyiegorov:
@copilot can you please errors?

PR #255, opened 2026-01-12
refactor(app): redesign convert to transfer bottom sheet with detached style

vitalyiegorov:
Replace the previous bottom sheet design with a detached, minimalistic style matching
the confirm action pattern. Changes include:
- Use dynamic sizing instead of fixed snap points
- Add transfer icon (ArrowRightLeft) instead of info icon
- Extract shared DETACHED_BOTTOM_SHEET_BORDER_PALETTE constant
- Update button text to "Convert to Transfer" for clarity
- Remove forwardRef in favor of React 19 native ref forwarding

### DIFF
--- a/.claude/agents/analyze.md
+++ b/.claude/agents/analyze.md
@@ -1,0 +1,218 @@
+---
+name: analyze
+description: "Comprehensive analysis of Budgie features, modules, or the entire monorepo"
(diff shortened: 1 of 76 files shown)


vitalyiegorov:
@copilot Please update all missing translations

PR #322, opened 2026-02-13
feat(app): add rules engine with suggest-rule UI and E2E tests

liaugust:
Add a rules engine supporting conditions (title, MCC code), actions (assign category, assign tag), and automatic categorization of bank-synced transactions. Include a suggest-rule modal triggered by an animated inline pill on transaction edit screens, rule form modals with modal-based selectors, and Maestro E2E flows for rule creation.


packages/app/src/rule/components/matching-rules-pill/matching-rules-pill.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const MatchingRulesPill = ({ matchingRulesCount }: Props) => {
+    const router = useRouter();
+
+    const handlePress = () => {
+        router.push('/settings/rules');
+    };
+
+    const label = matchingRulesCount === 1 ? t`${matchingRulesCount} matching rule` : t`${matchingRulesCount} matching rules`;
```
liaugust: why same translations in both cases ?

packages/app/src/rule/components/rule-suggestion-card/rule-suggestion-card.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }));
+
+    return {
+        enabled: true,
+        conditionMatchType: RuleConditionMatchTypeEnum.ALL,
+        conditions: [condition],
+        actions: [...categoryAction, ...tagActions],
+        applyToExisting: false
+    };
+};
```
liaugust, line 73: extract to separate util file

packages/app/src/rule/type/rule-condition-field.type.ts
```diff
@@ -0,0 +1 @@
+export type RuleConditionFieldType = 'field' | 'operator';
```
liaugust: why not enum?

packages/app/src/rule/type/rule-detection-mode.type.ts
```diff
@@ -0,0 +1 @@
+export type RuleDetectionModeType = 'suggest' | 'update' | 'match' | 'none';
```
liaugust: why not enum?

packages/app/src/rule/util/has-conflict-with-rule-outcomes.util.ts
```diff
@@ -0,0 +1,22 @@
...
+    readonly tagsChanged: boolean;
+}
+
+export const hasConflictWithRuleOutcomes = (params: HasConflictParams): boolean => {
+    const { userCategoryId, userTagIds, ruleOutcomes, categoryChanged, tagsChanged } = params;
+
+    const hasCategoryConflict = categoryChanged && isDefined(ruleOutcomes.categoryId) && userCategoryId !== ruleOutcomes.categoryId;
+
+    const userTagIdSet = new Set(userTagIds);
+    const hasTagConflict = tagsChanged && ruleOutcomes.tagIds.length > 0 && ruleOutcomes.tagIds.some(tagId => !userTagIdSet.has(tagId));
```
liaugust: isNotEmptyArray

packages/app/src/rule/util/compute-detection-mode.util.ts
```diff
@@ -0,0 +1,25 @@
+import { ComputeDetectionModeParamsInterface } from '../interface/compute-detection-mode-params.interface';
+import { RuleDetectionModeType } from '../type/rule-detection-mode.type';
+
+export const computeDetectionMode = (params: ComputeDetectionModeParamsInterface): RuleDetectionModeType => {
+    const { hasChanges, ruleCreated, isDismissed, matchingRulesCount, hasConflictWithMatchingRules } = params;
+    const canSuggest = hasChanges && !ruleCreated && !isDismissed;
+
+    if (matchingRulesCount === 1 && canSuggest) {
+        return 'update';
```
liaugust: why not enum?

packages/app/src/rule/util/compute-detection-mode.util.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    if (matchingRulesCount > 0 && !hasConflictWithMatchingRules) {
+        return 'match';
```
liaugust: why not enum?

packages/app/src/rule/util/compute-detection-mode.util.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const computeDetectionMode = (params: ComputeDetectionModeParamsInterface): RuleDetectionModeType => {
+    const { hasChanges, ruleCreated, isDismissed, matchingRulesCount, hasConflictWithMatchingRules } = params;
+    const canSuggest = hasChanges && !ruleCreated && !isDismissed;
+
+    if (matchingRulesCount === 1 && canSuggest) {
+        return 'update';
+    }
+
+    if (matchingRulesCount > 0 && !hasConflictWithMatchingRules) {
```
liaugust: isPositiveNumber

packages/app/src/rule/util/compute-detection-mode.util.ts
```diff
@@ -0,0 +1,25 @@
...
+        return 'match';
+    }
+
+    if (canSuggest) {
+        return 'suggest';
+    }
+
+    if (matchingRulesCount > 0) {
```
liaugust: isPositiveNumber

packages/app/src/app/rule-form.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { RuleFormCreate } from '../rule/components/rule-form-create/rule-form-create';
+import { RuleFormEdit } from '../rule/components/rule-form-edit/rule-form-edit';
+import { RuleFormResultType, useRuleFormModal } from '../rule/context/rule-form-modal.context';
+
+export default function RuleFormModal() {
+    const { currentParams, resolveRuleForm } = useRuleFormModal();
+
+    const handleSuccess = (result: RuleFormResultType) => {
+        resolveRuleForm(result);
+    };
```
liaugust: why do we need `handleSuccess` and not just `resolveRuleForm` ?

packages/app/src/rule/components/rule-summary-pills/rule-summary-pills.tsx
```diff
@@ -0,0 +1,53 @@
...
+                    const isLast = index === conditions.length - 1;
+                    const separator = isMatchAll ? <Trans>and</Trans> : <Trans>or</Trans>;
+
+                    return (
+                        <Fragment key={condition.id}>
+                            <RuleConditionPill condition={condition} />
+                            {isLast ? null : <Text className="text-xs text-secondary-foreground">{separator}</Text>}
+                        </Fragment>
+                    );
+                })}
```
liaugust: Extract everything in `map` callback into a separate component

packages/app/src/rule/components/swipeable-rule-card/swipeable-rule-card.tsx
```diff
@@ -0,0 +1,151 @@
...
+            </Animated.View>
+        );
+    }
+
+    return (
+        <GestureDetector gesture={panGesture}>
+            <Animated.View style={animatedStyle}>
+                <View
+                    testID={cardTestID}
+                    className={`flex-row items-center gap-sm px-lg py-sm bg-ghost-background rounded-xl shadow-sm ${cardClassName ?? 'self-start'}`}
```
liaugust: use `cva`

packages/app/src/rule/interface/rule-evaluation-input.interface.ts
```diff
@@ -0,0 +1,14 @@
+import { TransactionCreateInputInterface } from '@budgie/contracts';
+
+interface RuleEvaluationEntryInterface {
+    readonly type: TransactionCreateInputInterface['entries'][number]['type'];
+    readonly categoryId: TransactionCreateInputInterface['entries'][number]['categoryId'];
+    readonly accountId: TransactionCreateInputInterface['entries'][number]['accountId'];
+    readonly amount: TransactionCreateInputInterface['entries'][number]['amount'];
+    readonly mccCategoryId: TransactionCreateInputInterface['entries'][number]['mccCategoryId'];
```
liaugust: why do we type these fields using `TransactionCreateInputInterface['name'][number]['name']` syntax ?

packages/app/src/rule/util/build-rule-condition-sql.util.ts
```diff
@@ -0,0 +1,92 @@
...
+    const column = getColumnForField(condition.field);
+
+    if (!isDefined(column)) {
+        return null;
+    }
+
+    return buildOperatorSql(column, condition.operator, condition.value, condition.secondaryValue);
+};
+
+export type { RuleConditionInput };
```
liaugust: remove

packages/app/src/rule/util/evaluate-rule-condition.util.ts
```diff
@@ -0,0 +1,99 @@
...
+            return inValues.includes(String(fieldValue).toLowerCase());
+        }
+
+        default:
+            return false;
+    }
+};
+
+export const evaluateRuleCondition = (
+    condition: Pick<RuleConditionEntityInterface, 'field' | 'operator' | 'value' | 'secondaryValue'>,
```
liaugust: extract to separate interface and reuse where needed (I've seen this multiple times already)

packages/app/src/rule/util/has-matching-rule.util.ts
```diff
@@ -0,0 +1,37 @@
...
+export const doesRuleMatchTransaction = (
+    rule: RuleWithRelationsEntityInterface,
+    transactionInput: RuleEvaluationInputInterface,
+    suggestRuleData: SuggestRuleDataInterface
+): boolean => {
+    if (!isNotEmptyArray(rule.conditions)) {
+        return false;
+    }
+
+    const evaluator = rule.conditionMatchType === RuleConditionMatchTypeEnum.ANY ? 'some' : 'every';
```
liaugust: why not enum for `some` and `every` ?

packages/app/src/rule/util/select-suggest-condition.util.ts
```diff
@@ -0,0 +1,65 @@
+import { RuleConditionCreateInputInterface, RuleConditionFieldEnum, RuleConditionOperatorEnum } from '@budgie/contracts';
+
+import { isNotEmptyString } from '@rnw-community/shared';
+
+import { cleanMerchantTitle } from './clean-merchant-title.util';
+
+// eslint-disable-next-line lingui/no-unlocalized-strings
+const GENERIC_TITLE_PATTERNS = ['POS PURCHASE', 'POS', 'PURCHASE', 'PAYMENT', 'WITHDRAWAL', 'TRANSFER', 'DEBIT', 'CREDIT', 'ATM'];
```
liaugust, line 50: why do we need this ?

vitalyiegorov:
## SOTA Review: Overengineering & CLAUDE.md Compliance

**Scope:** 264 files, +11,452 / -1,463, 100 commits. Rule engine is functionally solid; the issues below are about **abstraction sprawl**, **schema mis-typing**, and a few **process regressions**. Not blocking — actionable as a polish pass.

### P0 — Block before merge

1. **AGENTS.md silently lost Maestro rules 7-12** (recovery branches, deeplink preference, system-prompt handling, settings outcome verification, `tapOn point` guidance). Unrelated to the feature; restore.
2. **`tx?: DB` should be `tx?: TX`** in `RuleRepository`, `RuleActionRepository`, `RuleConditionRepository` (`rule.repository.ts` L60/66/72/76 and the two siblings). `DB` defeats `(tx ?? this.db)` substitution and contradicts every other repo (`AccountRepository` etc.) — see `packages/contracts/CLAUDE.md` repo pattern.
3. **Single-member enum** `RuleConditionAssociationEnum { RULE = 'rule' }` (rule 45 violation). One member, one usage, exported from `index.ts`. Delete; inline `'rule'` in `rule-condition-entity.relations.ts`.
4. **Mutable runtime state in `*.constant.ts`** — `rule/constant/dismissed-suggestions.constant.ts` exports `new Set<string>()` as a mutable module-level singleton. Constants must be immutable (rule 29 spirit). Move to `useRef`/state inside `use-suggest-rule-detection.hook.ts`.

### P1 — Should fix this branch

5. **Single-consumer utils to inline** (rule 51). All 1-importer:
   - `find-all-matching-rules.util.ts` — body is `rules.filter(...)`. Delete; inline at hook.
   - `convert-transaction-to-transfer.util.ts` — only `RuleEngineService` uses it → make `private convertTransactionToTransfer()` (rule 43).
   - `escape-sql-like-value.util.ts` → fold into `build-rule-condition-sql.util.ts`.
   - `build-rule-condition-sql.util.ts` → fold into `build-rule-conditions-where.util.ts` (single SQL pipeline).
   - `build-rule-create-input.util.ts`, `find-duplicate-rule.util.ts` → inline into `rule-suggestion-card.tsx`.
   - `build-rule-input-from-prefill.util.ts` → inline into `use-rule-form.hook.ts`.
   - `clean-merchant-title.util.ts`, `extract-rule-action-outcomes.util.ts`, `has-conflict-with-rule-outcomes.util.ts`, `has-convertible-type-condition.util.ts` → inline into `use-suggest-rule-detection.hook.ts`.
   - `get-suggest-rule-field-value.util.ts` → inline into `has-matching-rule.util.ts`.

6. **Single-consumer constants to inline** (rules 39, 51). All 1-importer:
   - `field-operators.constant.ts`, `operator-options.constant.ts` → into `rule-condition-operator-selector.tsx`.
   - `rule-action-type-options.constant.ts` → into `rule-action-type-selector.tsx` (also: inline `RuleActionTypeOptionInterface` is rule-19 violation).
   - `rule-condition-match-type-labels.constant.ts` → into `rule-condition-match-type-selector.tsx`.

7. **Two near-identical relations interfaces** (rule 51) — `RuleWithRelationsEntityInterface` vs `RuleWithActionsRelationsEntityInterface` differ only in whether `RuleActionEntity` is "with-relations". Either widen actions to optional `category`/`tag` and collapse to one, or move the variance to the single repository method's return type.

8. **`applyToExisting: boolean` in `RuleCreateInputSchema`** (rule 37). Service/UI directive, never persisted. Move to an app-side form schema; `RuleUpdateInputSchema` correctly omits it — current asymmetry is a smell.

9. **Component pass-throughs to delete** (rule 51):
   - `rule-condition-mcc-selector` — 19-line `<FormItem>` wrapper around `*-content`. Inline the `FormItem` into the content component.
   - `rule-condition-external-source-selector`, `rule-condition-transaction-type-selector` — pure forwarders to `RuleConditionValueEnumSelector`. Inline into the `RuleConditionValueInput` switch.

10. **Inline interfaces above components** (rule 19) — `Option<T>` / `Props<T>` declared at top of `rule-action-bottom-sheet-selector.tsx`, `rule-condition-bottom-sheet-selector.tsx`, `rule-condition-value-enum-selector.tsx`. Move to `rule/interface/`.

11. **Hook with no encapsulation** (rule 47) — `use-rule-condition-value-field.hook.ts` is a 2-line wrapper around `useFormContext` + `useController`. Inline at its 2 consumers.

12. **`importer.service.ts` positional coupling** — `ruleEngineService.applyRulesToTransactions(createdTransactions.map(t => t.id), transactions)` assumes `bulkCreate` returns rows in input order. Pair explicitly via `Map<id, input>` or document the ordering guarantee on `bulkCreate`.

### P2 — Nice to have

13. **Missing FK indexes** in `0020_add_rules.sql` on `rule_conditions.rule_id`, `rule_actions.rule_id`. Every read filters by `rule_id`; SQLite doesn't auto-index FKs.
14. **`compute-detection-mode-params.interface.ts`** — 5-field params interface, single consumer; pass fields directly or inline as a local type.
15. **Module-level `FIELD_OPTIONS` inside component file** — `rule-condition-field-selector.tsx` defines `FIELD_OPTIONS` at module scope; move to `rule/constant/` per rule 10 (or inline if kept).

### What's already good

- 3-service split (`rule.service` / `rule-matcher.service` / `rule-engine.service`) is justified — each has multi-method cohesion and shared private helpers (rule 38).
- No `useEffect` boot hook for the engine — clean stateless orchestrator (rules 46-47 respected).
- Modal-provider nesting follows the established pattern (no new pollution).
- `app.service.truncateData` got `@Log` — net improvement.
- Field naming, repository constructor injection, soft-delete via `archiveById` all match existing conventions.

### Bottom line

The engine itself is well-architected. The drag is **abstraction sprawl** in `rule/util/`, `rule/constant/`, and a handful of selector components — about **15 files worth deleting/inlining** without losing any functionality, plus the 4 P0s. Doing the P0+P1 sweep would shrink the rule module footprint by ~10-15% and bring it in line with the project's single-consumer rule.

vitalyiegorov:
## SOTA Review Addendum: keep the engine, shrink the scaffolding

I read the latest SOTA review comment and rechecked the current `feat/rules` diff. I agree with the existing P0/P1 direction: the core engine split is defensible, but the branch is carrying too much scaffolding around it. The best cleanup pass is still P0 + P1 from the previous comment: restore removed AGENTS/Maestro rules, fix repository transaction typing against the contracts package guidance, remove the single-member relation enum, move dismissed suggestion state out of a `*.constant.ts`, and inline the single-consumer utility/constant/pass-through layer.

Additional findings:

1. **Form state can reset during editing.** `useRuleForm()` passes `values: defaultValues` into RHF, while `RuleFormEdit` rebuilds `defaultValues` as a fresh object from `rule` on every render. Any rerender can push the original server values back over the user’s in-progress edits. Prefer stable `defaultValues`, then call `reset()` only when the loaded `rule.id`/payload actually changes.

2. **Automatic rule application swallows failures.** `applyRulesToTransactions()` uses `Promise.allSettled()` and ignores rejected results, so import can finish as if rules applied even when every application failed. This is especially risky for `ADD_TAG`: `transaction_tags` has a `(transaction_id, tag_id)` primary key, so adding an already-present tag throws, rolls back that transaction’s rule changes, and the caller never sees it. Either count/report failures like `applyRuleToMatchingTransactions()` or make idempotent actions idempotent before swallowing anything.

3. **`applyToExisting` is a dead cross-layer field.** It lives in `RuleCreateInputSchema`, default values, edit defaults, and prefill data, but `confirmApplyToExisting()` never reads it and `ruleService.create()` drops it. This is exactly the kind of UI directive that should not be in contracts. Delete it or make it an app-side form-only value with an actual consumer.

4. **UI-only invariants need service/schema backing.** `CONVERT_TO_TRANSFER` is only hidden by selector filtering when the conditions include an expense/income type condition. If conditions change after selecting it, the value can remain. `convertTransactionToTransfer()` then silently no-ops for non-expense/income, while the rule flow can still mark the transaction as rule-updated. Enforce the invariant in form validation or service input, not only the option list.

Bottom line: keep `rule.service`, `rule-matcher.service`, and `rule-engine.service`, but delete the one-importer utility/constant layer and move invariants closer to the data boundary. That will make this feature feel like the rest of the repo instead of a generated mini-framework.

packages/app/src/rule/hooks/use-rule-form.hook.ts
```diff
@@ -0,0 +1,187 @@
...
+        if (!hasConditions) {
+            return false;
+        }
+
+        const count = await ruleMatcherService.countMatchingTransactions({
+            conditions: values.conditions,
+            conditionMatchType: values.conditionMatchType
+        });
+
+        if (count <= 0) {
```
liaugust: !isPositiveNumber

packages/app/src/rule/components/swipeable-rule-card/swipeable-rule-card.tsx
```diff
@@ -0,0 +1,189 @@
...
+    };
+
+    if (status === 'success') {
+        return (
+            <Animated.View entering={FadeIn.duration(200)} exiting={FadeOut.duration(300)}>
+                <View className="flex-row items-center gap-xs px-lg py-sm bg-ghost-background rounded-xl shadow-sm">
+                    <Icon icon={UserIconNameEnum.CircleCheck} size={CHECK_ICON_SIZE} className="text-positive-foreground" />
+                    <Text className="text-xs text-secondary-foreground font-medium">{successMessage(appliedCount)}</Text>
+                </View>
+            </Animated.View>
```
liaugust: Extract to separate component and reuse here and below

packages/app/src/rule/hooks/use-suggest-rule-detection.hook.ts
liaugust, line 1: split by multiple smaller files. it is hard to read this one

packages/app/src/rule/interface/rule-prefill-data.interface.ts
```diff
@@ -0,0 +1,12 @@
+import { RuleConditionFieldEnum } from '@budgie/contracts';
+
+type RulePrefillConditionInterface = {
+    readonly field: RuleConditionFieldEnum;
+    readonly value: string;
+};
```
liaugust: extract to separate file

packages/app/src/rule/service/rule-engine.service.ts
```diff
@@ -0,0 +1,411 @@
...
+                        return previousConverted || converted;
+                    }),
+                Promise.resolve(false)
+            );
+            await transactionRepository.updateById(transactionId, { updatedBy: TransactionUpdatedByEnum.RULE }, transaction);
+
+            if (convertedToTransfer) {
+                await accountBalanceIncrementalService.updateAllBalances(true, transaction);
+            }
+        });
```
liaugust, line 163: it is not readable. please improve

PR #346, opened 2026-03-21
fix: prevent inactive-state embedding generation after sync imports

vitalyiegorov:
## Problem
- sync and import flows can create transactions while the app is not active
- embedding generation could still run in that state and hit the same failure path again when the app returns

## Investigation
- inline embedding generation was not guarded by app activity
- missing embedding backfill could start as soon as embeddings became ready, even if the app was still inactive
- batch processing still relied on loading existing context keys instead of querying only the missing merchant and comment contexts

## Fix
- add an app-state-aware embedding backfill service that runs only when embeddings are ready and the app is active, and retries when the app returns to the foreground
- stop inline transaction embedding generation as soon as the app is no longer active
- switch merchant and comment batch processing to repository queries that fetch only missing contexts
- remove leftover `LlmProvider` debug logging

## Validation
- attempted `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd`
- `yarn ts` currently fails on `main` in `packages/app/src/@generic/hook/use-bootstrap-reset.hook.ts` with pre-existing TS4111 index-signature property-access errors unrelated to this PR


vitalyiegorov:
Closing this oversized branch in favor of #359. The crash-fix direction is useful, but this PR is too polluted to merge safely as-is. I’m keeping the clean Maestro/E2E branch as the integration point and will port any crash-fix slice there separately if needed.

PR #363, opened 2026-04-10
fix(bank-sync): parse inline Abschlussbuchung transactions in Erste modern PDF

vitalyiegorov:
## Summary
- Erste Bank changed their PDF statement format to include inline transactions in the Abschlussbuchung (closing entries) section
- Lines like `Habenzinsen 31.03.2026 0,33` have description+date+amount on one line, which the existing regex missed (it required date at line start)
- Added `ERSTE_MODERN_INLINE_TRANSACTION_REGEX` to match inline pattern and `createInlineTransaction` method to extract description from the line
- Previously lost transactions: Habenzinsen, Kest, Kontoführung, Bonus Kontoführung — now all parsed correctly
- Also fixes corrupted descriptions on the transaction following the Abschlussbuchung section

## Test plan
- [ ] Import PDF 008 (with Abschlussbuchung section) and verify all 11 transactions appear with correct titles
- [ ] Import PDF 009 (standard format without inline transactions) and verify 5 transactions — regression check
- [ ] Verify inline credit transactions (Habenzinsen, Bonus Kontoführung) have positive amounts
- [ ] Verify inline debit transactions (Kest, Kontoführung) have negative amounts
- [ ] Verify Polizze Nr. transaction after inline section has correct title (Wiener Städtische, not Abschlussbuchung garbage)


packages/app/src/app/(main)/pin.tsx
```diff
@@ -70,7 +70,7 @@ export default function PinScreen() {
         }
     };
 
-    const handleBiometricAuth = async () => {
+    const handleBiometricAuth = useCallback(async () => {
```
liaugust: Are you sure we need it?

liaugust (changes requested):

PR #368, opened 2026-04-14
fix(app): prevent AI providers from running when AI is disabled

vitalyiegorov:
## Summary

- Add `aiEnabled` config flag to `app.config.js` driven by `EXPO_PUBLIC_AI_DISABLE` env var and E2E variant
- Create `isAiEnabled()` utility that checks the config flag
- Add `AiStatusDisabledProvider` and `AiEmbeddingProgressDisabledProvider` — lightweight no-op providers
- Make all three AI providers (`LlmProvider`, `AiStatusProvider`, `AiEmbeddingProgressProvider`) conditional — only mount when AI is enabled
- Remove leftover debug `console.log` statements from `LlmProvider`

## Why

`AiStatusProvider` and `AiEmbeddingProgressProvider` were always mounted unconditionally in the root layout, even when AI was disabled via `EXPO_PUBLIC_AI_DISABLE=true`. This caused background crashes from AI subsystems initializing and running when they shouldn't be.

## Test plan

- [ ] Run app with `EXPO_PUBLIC_AI_DISABLE=true` — verify no AI-related background activity or crashes
- [ ] Run app normally (AI enabled) — verify AI features work as before
- [ ] Run `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd`

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -21,6 +21,7 @@
         "packages/contracts/src/**/*.enum.ts",
         "packages/app/src/category/enum/system-category-id.enum.ts",
         "packages/app/src/ai/provider/llm.provider.tsx",
+        "packages/app/src/@generic/hook/use-foreground-recovery.hook.ts",
         "packages/app/src/ai/hook/use-llama-llm.hook.ts",
         "packages/app/src/ai/context/llm.context.ts"
     ],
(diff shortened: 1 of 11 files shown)


knip.json
```diff
@@ -21,6 +21,7 @@
         "packages/contracts/src/**/*.enum.ts",
         "packages/app/src/category/enum/system-category-id.enum.ts",
         "packages/app/src/ai/provider/llm.provider.tsx",
+        "packages/app/src/@generic/hook/use-foreground-recovery.hook.ts",
```
liaugust, line 24: Why do we need to ignore it?

liaugust (changes requested):

liaugust (approved):

PR #374, opened 2026-04-17
refactor(app): unify AI providers and move embedding to foreground drainer

vitalyiegorov:
## Summary

Replaces the dual-tree AI stack (`ConditionalLlmProvider` + `DisabledAiStackProvider` + `React.lazy(LlmProvider)`) with a single `AiProvider` + `AiModeEnum` that enforces "no Metal GPU calls outside foreground" at the native-call boundary. Also bumps Expo SDK 55 patch-latest, drizzle-orm to 0.45, enforces braces on all control-flow bodies, and compacts the date filter sheet.

### AI refactor (primary)

- New `AiModeEnum`: `Disabled | Initializing | Ready | Suspended | Error`
- Single `AiProvider` owns the llama.rn lifecycle, AppState subscription, 30s background release debounce, and exposes `stt` alongside `llm`
- `isNativeCallSafe(mode)` util gates every native call site (provider, drainer, data-prep loops)
- `needsEmbedding` column on transactions (+ backfill migration 0016); producers flag rows, a foreground-only `EmbeddingDrainerService` drains the queue — bank sync / transaction create no longer call Metal directly
- Brain indicator / suggestion hooks read `mode`, eliminating silent-disable paths
- Debug ring buffer for mode transitions (`aiDebugBuffer` + `useAiDebug()`)
- 14 obsolete files deleted (old providers, contexts, `useLlamaLlm`, `useForegroundRecovery`)

### Other

- **Expo 55.0.4 → 55.0.15** across all `expo-*` packages; react-native 0.83.2 → 0.83.4
- **drizzle-orm 0.44.7 → 0.45.2** (app + contracts), drizzle-kit 0.31.10
- **ESLint `curly: all` + `nonblock-statement-body-position: below`** enforced; rule 26 in AGENTS.md
- Removed unused `@libsql/client` (docs confirm it's not needed for `driver: 'expo'`)
- `DevMenuController` conditionally rendered `__DEV__ && <DevMenuController />` so its JSX is tree-shaken from prod bundles
- Compact date filter sheet: quick-period chips moved into the drawer right above the apply button

### Spec & plan

Design doc and implementation plan live under `docs/plans/` (gitignored):
- `2026-04-17-ai-providers-transactions-design.md`
- `2026-04-17-ai-providers-transactions-plan.md`

## Test plan

- [x] `yarn format && yarn ts && yarn lint && yarn cpd && yarn test` all pass
- [x] iOS simulator (iPhone 16 Pro) ad-hoc build succeeds with new unified AiProvider mounted
- [ ] **Golden path** — open transaction create screen, confirm brain animates during init, suggestions appear once `mode === Ready`
- [ ] **Embedding queue** — create a transaction, DB row `needs_embedding=1`, drainer clears to 0 within ~5s on foreground
- [ ] **Background safety** — background app for >30s, return; verify no Metal crash, mode goes Suspended → Initializing → Ready
- [ ] **Error state** — block model download, confirm `mode=Error` and `retry()` wiring
- [ ] **E2E** — `APP_VARIANT=e2e` renders all transaction pages with `mode=Disabled`, no brain / suggestions / crashes
- [ ] **Date filter sheet** — pills sit above Apply button, sheet height reduced, all four periods still switch correctly


vitalyiegorov:
# SOTA Review — PR #374

Five parallel domain reviewers analyzed the AI provider refactor, data layer, UI changes, composition, and Expo upgrade paths. Findings consolidated and de-duplicated below.

## What's done well

- `AiModeEnum` cleanly replaces 3–4 booleans; enum-over-boolean is the right call.
- `isNativeCallSafe(mode)` double-gates (`mode === Ready` + `AppState === 'active'`) at drain start AND per-row iteration. Foreground-only invariant holds.
- `cancelledRef` interlock in init effect (checks between `await initLlama` calls, releases contexts on cancel) is thorough.
- `merchantEmbeddingRepository.upsert` + `commentEmbeddingRepository.upsert` use `onConflictDoUpdate` with correct composite targets — no duplicate vectors.
- `isNull(deletedAt)` in `findPendingEmbedding` handles orphan-row scenario correctly.
- `useMigrations` blocks render until m0016 completes, so OTA-update path is safe for existing users.
- `@libsql/client` removal is clean — drizzle-orm 0.45 marks it optional peer.
- `AiBrainProgress` animation choreography (`withRepeat`/`withSequence`) + `SuggestionRowLayout` FadeIn/FadeOut are well implemented.

---

## P0 — Block merge

### 1. Context value + `llm` object rebuilt every render → re-render avalanche
**File:** `packages/app/src/ai/provider/ai.provider.tsx:380-393`

Both `const llm: LlmInterface = { ... }` (lines 380–391) and `const value = { mode, llm, stt, progress, isEmbedding, downloadProgress, retry, refreshProgress }` (line 393) are object literals reconstructed every render. During model download, `downloadProgress` ticks on every progress callback; `progressVersion` bumps every 2s from the drainer. Every `useAi()` consumer re-renders on every tick — even components reading only `mode` or `llm`.

React 19 Compiler can't memoize object literals with mixed-stability inputs. `CLAUDE.md` rule 1 ("no manual memoization") applies to leaf components; **providers with mixed-stability fan-out are the canonical exception**.

**Fix:** Wrap `llm` and `value` in `useMemo`, OR split context into `AiCapabilityContext` (stable: `generate/embedding/batchEmbedding/interrupt/retry`) and `AiStatusContext` (volatile: `mode/progress/isEmbedding/downloadProgress`). Consumers like `use-comment-suggestion` that only need capabilities avoid the download-progress re-render cascade.

### 2. `Error` mode has no visible UI surface
**File:** `packages/app/src/transaction/components/suggestion-loading-indicator/suggestion-loading-indicator.tsx:22`

```ts
const statusLabel = mode === AiModeEnum.Initializing ? t`Loading AI model...` : '';
```

Every mode other than `Initializing` returns empty string. When `mode === Error`, user sees a pulsing brain with no message, no affordance, no retry tap target. The spec called for an explicit error surface; this misses the entire point of adding the `Error` enum value.

**Fix:** Add `mode === Error` branch to `statusLabel` with `t\`AI unavailable – tap to retry\`` and wire the existing `handleBrainPress` to call `retry()` from `useAi()` when mode is Error.

---

## P1 — Before merge

### 3. Silent embedding discard when `embed()` returns `[]`
**File:** `packages/app/src/ai/service/embedding-drainer.service.ts` (processRow, ~lines 138–170)

If `embed()` returns empty array, neither `upsert` nor `replaceTags` fires — but `await transactionRepository.updateById(row.id, { needsEmbedding: false })` at the bottom runs unconditionally, clearing the flag as if processed. Per spec: failures should leave `needsEmbedding: true` for retry.

**Fix:** Track a `stored` boolean inside each branch; only call the clearing `updateById` when `stored === true` OR when `categoryId` is not positive (the intentional skip case).

### 4. Redundant `markForEmbedding` on create pages
**Files:** `packages/app/src/app/(main)/create-transaction/{expense,income,transfer}.tsx`

`transactionService.bulkCreate` (transaction.service.ts:76) already stamps `needsEmbedding: true` on every new row. Create screens then call `markForEmbedding({ transactionId: result.id })` which fires a redundant `updateById`. In the race window between INSERT and this second write, the drainer may process the row, clear the flag, then `markForEmbedding` sets it again → drainer embeds twice.

**Fix:** Remove `markForEmbedding` from create screens. Keep it on **update** screens (where the user edits a transaction and its embedding must be recomputed). On create screens, replace with a plain `refreshProgress()` call from `useAi()`.

### 5. Drainer polls every 2s even when queue is empty
**File:** `packages/app/src/ai/service/embedding-drainer.service.ts:85-112`

`finally` block calls `scheduleDrain()` unconditionally when mode is Ready + app active. When `rows.length === 0`, the drain exits early but `finally` still reschedules → infinite 2s SQLite scans forever, even after backfill completes and no transactions need embedding.

**Fix:** Track `hadWork = rows.length > 0` before the early return; only call `scheduleDrain()` in `finally` when `hadWork === true`. `handleAppState` and `refreshProgress` (from consumer writes) remain the wake mechanisms.

### 6. No index on `needs_embedding`
**Files:** `packages/contracts/src/transaction/table/transaction-entity.table.ts`, `packages/app/drizzle/0016_add_needs_embedding.sql`

`findPendingEmbedding` filters on `needs_embedding = 1 AND deleted_at IS NULL` with no supporting index. Every 2s SQLite scan is O(n). Fix combines with #5.

**Fix:** Add to migration 0016:
```sql
--> statement-breakpoint
CREATE INDEX idx_txn_pending_embedding ON transactions(id) WHERE needs_embedding = 1 AND deleted_at IS NULL;
```
Also mirror in the Drizzle schema via `index(...).on(...).where(...)`.

### 7. `Error` mode doesn't recover on foreground
**File:** `packages/app/src/ai/provider/ai.provider.tsx:234-265`

`handleChange` only re-inits when `modeRef.current === AiModeEnum.Suspended`. If init threw and the user backgrounds + foregrounds quickly (before the 30s release timer fires), mode is still `Error`, and returning to foreground does nothing. User is stuck in silent-broken state until they manually tap retry.

**Fix:** Treat `Error` as recoverable on foreground — extend the condition to `[Suspended, Error].includes(modeRef.current)`.

### 8. `EMBEDDING_COMPLETENESS_THRESHOLD = 90` duplicated
**Files:** `packages/app/src/transaction/components/suggestion-row-layout/suggestion-row-layout.tsx:17` and `packages/app/src/ai/hook/use-suggestion-base.hook.ts:26`

Two independent declarations of the same constant. Per `CLAUDE.md` rule 10 (constants in `/constant` folder).

**Fix:** Extract to `packages/app/src/ai/constant/embedding-completeness-threshold.constant.ts`; import both sites.

### 9. `useAi()` is an identity wrapper — violates rule 16
**File:** `packages/app/src/ai/hook/use-ai.hook.ts:3`

```ts
export const useAi = () => useAiContext();
```

Rule 16 ("no redundant wrapper functions") applies directly.

**Fix:** Delete `use-ai.hook.ts`; replace all `useAi()` imports with `useAiContext()` from `../context/ai.context`.

### 10. `handleMarkError` empty-body named function violates rules 3/6/16
**File:** `packages/app/src/ai/hook/use-embedding-generator.hook.ts:17`

```ts
const handleMarkError = (): void => {
    // intentionally empty...
};
```

Rule 3 (no comments), rule 6 (use `emptyFn` from `@rnw-community/shared`), rule 16 (no redundant wrappers).

**Fix:** Replace both call chains with `.catch(emptyFn)`.

### 11. Drainer registers its own `AppState` listener
**File:** `packages/app/src/ai/service/embedding-drainer.service.ts:36`

The provider already owns an `AppState` listener (ai.provider.tsx:265). Two independent subscriptions for tightly coupled lifecycle is a design smell and doubles the chance of races between them.

**Fix:** Remove the drainer's internal listener; have the provider's existing listener call `embeddingDrainerService.pause()` / `.resume()` on state change.

### 12. `start()` doesn't guard against double-invocation
**File:** `packages/app/src/ai/service/embedding-drainer.service.ts:33-37`

If `mode` flickers `Ready → Suspended → Ready`, `start()` may be called twice without a `stop()` between if React effect sequencing ever misbehaves.

**Fix:** First line of `start()`: `this.stop();`.

### 13. `expo` uses `^`, all other `expo-*` use `~`
**File:** `packages/app/package.json:41`

```json
"expo": "^55.0.15",
```

Every sibling `expo-*` package uses `~` (patch-locked). `^55.0.15` allows minor bumps on next `yarn install`, while `expo-sqlite: ~55.0.15` will not — a latent incompatibility trap.

**Fix:** `"expo": "~55.0.15"`.

### 14. 17 leftover `console.log` in `use-ai-data-preparation.hook.ts`
**File:** `packages/app/src/ai/hook/use-ai-data-preparation.hook.ts` (lines 58, 61, 66, 72, 80, 90, 113, 120, 122, 128, 135, 137, 144, 155, 166, 172)

Every one has `// eslint-disable-line no-console` suppression.

**Fix:** Delete all 17. This refactor introduced `aiDebugBuffer` — replace any genuinely useful log with `aiDebugBuffer.push(...)` and drop the rest.

### 15. Chip horizontal scroll bleeds into drawer padding on small devices
**File:** `packages/app/src/app/date-filter.tsx:46` and `packages/app/src/@generic/component/filter-sheet/filter-sheet-drawer/filter-sheet-drawer.tsx:24`

Drawer applies `px-xl` to children. Chip `ScrollView` is indented by `xl` and starts non-flush. iOS convention for bottom-sheet chip rows is edge-bleed.

**Fix:** Add `-mx-xl` to the ScrollView className, and `contentContainerClassName="gap-x-sm px-xl"` to restore the first-chip padding but let overflow bleed.

### 16. `Suspended` flash on app-switcher return
**File:** `packages/app/src/transaction/components/suggestion-row-layout/suggestion-row-layout.tsx:25-26`

`showBrain` evaluates true whenever `progress < 90`, regardless of `mode`. On slow foreground returns the brain re-enters with `FadeIn.delay(400)` twice during `Suspended → Initializing → Ready`.

**Fix:** Gate `showBrain` with `mode !== AiModeEnum.Suspended`.

---

## P2 — Follow-up

### 17. `generate` mutex dual-branch duplication — `ai.provider.tsx:298-304`
The `.then(onFulfilled, onRejected)` has identical bodies. Simplify with `.catch(() => {})` on `previous`, then a single `.then(() => runCompletion(...))`.

### 18. Backfill throughput for large datasets — `0016_add_needs_embedding.sql`
`BATCH_SIZE=10, DRAIN_INTERVAL_MS=2000` → ~5 rows/sec. A user with 5,000 transactions spends ~17 minutes draining. Consider detecting "backfill mode" (`pending / total > 0.5`) and using `BATCH_SIZE=25, DRAIN_INTERVAL_MS=500` until drained.

### 19. `processRow` doesn't guard missing `entry` — `embedding-drainer.service.ts:124`
`const [entry] = row.entries` — if entries is empty, `entry.categoryId` throws, caught by outer `catch {}`, leaves flag `true`, retried forever.

**Fix:** `if (!isDefined(entry)) { await transactionRepository.updateById(row.id, { needsEmbedding: false }); return; }`

### 20. `AiContextInterface` mixes unrelated concerns
`llm + stt + mode + progress + isEmbedding + downloadProgress + retry + refreshProgress`. Splitting into 3 sibling contexts exposed by the same provider (capability / status / telemetry) would reduce re-render surface. Linked to P0 #1.

### 21. Date filter `pt-md` may feel cramped on iPhone SE — `date-filter.tsx:39`
Consider `pt-lg` as a middle ground between `pt-4xl` (old) and `pt-md` (new).

### 22. `isEmbedding` is redundant with `progress` — `ai.provider.tsx:95,135`
`isEmbedding === (progress > 0 && progress < 100)`. Can be derived at call sites; removing the state reduces the update surface.

### 23. Apply-button label switch without crossfade — `filter-sheet-apply.tsx`
"Show all dates" / "Show selected range" hard-cuts on selection change. iOS convention crossfades.

### 24. `expo-font` (~55.0.6) and `expo-image` (~55.0.8) lag behind rest of expo-* cluster
Minor but inconsistent.

---

## Summary

**Recommendation:** Request changes.

P0 #1 (re-render avalanche) and P0 #2 (invisible Error state) are both regressions of the stated goals — the whole point of the refactor was to eliminate silent-disable paths and prevent unnecessary work, yet the current implementation has both. P1 #3–#5 (silent discard, redundant flag, empty-queue poll) are consistency holes in the background-safety guarantees. Everything else is polish.

Fix P0 + P1 #3–#7, then this is merge-ready.

packages/ai/src/embedding/service/embedding-suggestion.service.ts
```diff
@@ -122,14 +144,36 @@ export class EmbeddingSuggestionService {
...
+
+    private async prepareSuggestion(
+        transactionTitle: string,
+        mccDescription: string | null,
+        comment: string,
+        aiContext: string
+    ): Promise<{
+        readonly resolved: { readonly serialized: Uint8Array; readonly distanceThreshold: number };
+        readonly methodStart: number;
+    } | null> {
```
liaugust: Extract to interface

packages/app/scripts/after.txt
liaugust, line 1: remove? tsignore?

packages/app/src/ai/component/ai-system-status-banner/ai-system-status-banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+        return null;
+    }
+
+    const icon = BANNER_ICON[umbrella.state];
+    const colorClass = BANNER_COLOR[umbrella.state];
+    const showPercent = umbrella.state === AiSystemUmbrellaStateEnum.Downloading && umbrella.downloadPercent > 0;
+
+    return (
+        <View className="flex-row items-center gap-x-sm px-lg py-md bg-secondary-background rounded-xl">
+            {icon !== null && <Icon icon={icon} size={ICON_SIZE} className={colorClass} />}
```
liaugust: isDefined

packages/app/src/ai/component/ai-system-status-banner/ai-system-status-banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+            <Text className={`flex-1 text-xs font-medium ${colorClass}`} numberOfLines={1}>
+                {umbrella.statusText}
+            </Text>
+            {showPercent && <Text className={`text-xs font-medium ${colorClass}`}>{`${umbrella.downloadPercent}%`}</Text>}
```
liaugust: cn

packages/app/src/ai/component/ai-system-status-banner/ai-system-status-banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const umbrella = useAiSystemUmbrella();
+    const isHealthy = umbrella.state === AiSystemUmbrellaStateEnum.Healthy;
+
+    if (isHealthy) {
+        return null;
+    }
+
+    const icon = BANNER_ICON[umbrella.state];
+    const colorClass = BANNER_COLOR[umbrella.state];
+    const showPercent = umbrella.state === AiSystemUmbrellaStateEnum.Downloading && umbrella.downloadPercent > 0;
```
liaugust: isPositiveNumber

packages/app/src/ai/enum/ai-subsystem-card-state.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/enum/ai-subsystem-status.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/hook/use-ai-subsystem-icon-animations.hook.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const useAiSubsystemIconAnimations = ({ percent, iconSize, pulsePeriodMs }: Params) => {
+    const reducedMotion = useReducedMotion();
+    const pulseValue = useSharedValue(FULL_OPACITY);
+    const fillHeight = useSharedValue((percent / PERCENT_DIVISOR) * iconSize);
+
+    useEffect(() => {
+        cancelAnimation(pulseValue);
+        if (reducedMotion || pulsePeriodMs === null) {
```
liaugust: isDefiend

packages/app/src/ai/hook/use-ai-subsystem-icon-animations.hook.ts
liaugust, line 1: one hook per file - split into two separate hook files

packages/app/src/ai/interface/ai-embedding-status-snapshot.interface.ts
liaugust, line 1: remove file. import AiSubsystemStatusSnapshotInterface directly from ai-subsystem-status-snapshot.interface.ts

packages/app/src/ai/interface/ai-translation-status-snapshot.interface.ts
liaugust, line 1: remove file

packages/app/src/ai/interface/embedding-snapshot.interface.ts
liaugust, line 1: why do we need such a re-exports?

packages/app/src/ai/service/ai-system-status.service.ts
```diff
@@ -0,0 +1,330 @@
...
+        const promises: Promise<void>[] = [];
+        if (chatService.getSnapshot().errorMessage !== null) {
+            promises.push(chatService.retry());
+        }
+        if (embeddingService.getSnapshot().errorMessage !== null) {
+            promises.push(embeddingService.retry());
+        }
+        if (sttService.getSnapshot().errorMessage !== null) {
+            promises.push(sttService.retry());
+        }
```
liaugust: isNotEmptyString

packages/app/src/ai/service/ai-system-status.service.ts
```diff
@@ -0,0 +1,330 @@
...
+
+        const translationBoosting = translationDrainerService.getSnapshot().state === DrainerStateEnum.Boosting;
+        const embeddingBoosting = embeddingDrainerService.getSnapshot().state === DrainerStateEnum.Boosting;
+        if (translationBoosting || embeddingBoosting) {
+            return this.deriveBoosting(translationBoosting, translationPending, embeddingPending);
+        }
+
+        if (translationPending > 0) {
+            const { total } = translationProgressStore.getSnapshot();
+            const trailer = embeddingPending > 0 ? t` • ${embeddingPending} tx queued` : '';
```
liaugust: isPositiveNumber

liaugust (changes requested):

vitalyiegorov:
@liaugust ready for re-review.

The six reviewer-flagged invariant leaks are addressed across 10 commits on top of `56087b632`. HEAD is `37bdc8e68`.

| Issue | Fix commits |
|---|---|
| P0 — eager STT init at tab mount | `61505520e`, `6f82b9ecf`, `37bdc8e68` — overlay mount gated on `isAiAvailable && isVoiceInputOpen`; overlay simplified to mount-as-intent with `voiceInputRef` unmount cleanup |
| P0 — suspended AI reported as healthy | `35ea650bf`, `1b94ea45d` — `AiSystemUmbrellaStateEnum` + `AiSystemStateEnum` gain `Suspended`/`Idle`; both derive waterfalls branch on them before the Ready/Healthy fallthrough; exhaustive `Record<Enum,…>` maps updated; i18n added for de/es/fr/uk |
| P1 — imported updates don't restamp `needsEmbedding` | `19f14215e` — new `deriveEmbeddingFlag()` in `@budgie/contracts` applied inside `transactionRepository.updateById`. Semantic-field update → restamp; transition to `TRANSFER`/`ADJUSTMENT` → clear. Explicit caller overrides still win. |
| P1 — cleanup queries fire without being awaited | `19f14215e`, `78237ec1e` — `clearResidueOnce` wraps the trio in `transactionAsync` for atomicity. Note: `runner.run(...)` on expo-sqlite is sync (`BaseSQLiteDatabase<'sync', …>`), so no `await` is added to the inner calls — ordering/exclusivity is provided by the wrapping transaction. |
| P1 — `TransactionEntitySchema` lies about `needsEmbedding` | `eb742a11d`, `d7d82d85c` — select schema non-optional; create schema keeps optionality via `.extend({ needsEmbedding: boolean().optional() })`. |
| P2 — `convertToTransfer` leaves flag stale | covered transitively by the repository-level `deriveEmbeddingFlag` — no caller-side belt-and-suspenders. |

Additional refinements surfaced during review loops:
- `refactor(app): align status derivation helpers for consistency` — matches the `const statuses = […] as const` pattern between umbrella + system services; `SuspendedOrIdleState` alias for the narrow return type.
- `fix(app): avoid double-cancel on VoiceInputOverlay close` — unmount cleanup is now the single cancel path.
- `chore(app): drop unused useIsAiReady hook, sync i18n …` — YAGNI; reintroduce when a real tap-affordance migration needs it.

Verification: `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd` all clean. Code quality / CodeQL / iOS build / EAS Deploy / Vercel / claude-review all pass in CI. The `E2E Tests on iOS` failure on `01.account-bank.flow (TabBar.Home is visible)` is pre-existing — the same assertion fails on the parent commit `56087b632` before this push.

Out-of-scope follow-ups intentionally not addressed here: UX polish (100ms VoiceInputOverlay exit fade), SQL perf beyond the atomicity fix, dependency churn.

packages/contracts/src/transaction/repository/transaction-pattern.repository.ts
```diff
@@ -226,210 +222,262 @@ export class TransactionPatternRepository {
...
-            .innerJoin(AccountEntityTable, ACCOUNT_JOIN_CONDITION)
-            .leftJoin(CategoryEntityTable, CATEGORY_JOIN_CONDITION)
-            .leftJoin(MccCategoryEntityTable, MCC_CATEGORY_JOIN_CONDITION)
-            .where(and(...conditions))
-            .groupBy(TransactionEntityTable.title, AccountEntityTable.id)
-            .having(this.buildMonthlyHavingConditions(distinctMonth, modeDayCount))
-            .orderBy(desc(sql`COUNT(DISTINCT ${distinctMonth})`), desc(sql`MAX(${TransactionEntityTable.operatedAt})`));
+        const { entryType, recencyTimestamp, tzOffset, type, defaultInstrumentId, displayMonth } = this.buildMonthlyQueryContext(query);
+
+        const bankSql = `
```
liaugust, line 256: Don't you think it is way too big and not that readable? I think you can extract some sub-selects into separate constants

packages/contracts/src/tag/repository/tag.repository.ts
```diff
@@ -22,8 +32,19 @@ export class TagRepository {
     }
 
     findBySearchQuery(search: string) {
+        const trimmed = search.trim();
+        if (trimmed === '') {
```
liaugust: isEmptyString

packages/contracts/src/category/repository/category.repository.ts
```diff
@@ -12,25 +13,50 @@ import type * as schema from '../../schema';
...
     }
 
     findBySearchQuery(search: string, includeDefault: boolean) {
-        const searchQuery = like(CategoryEntityTable.titleSearch, `%${search.toLowerCase()}%`);
+        const trimmed = search.trim();
+        const baseFilter = includeDefault
+            ? eq(CategoryEntityTable.isSystemCategory, false)
+            : and(eq(CategoryEntityTable.isDefault, false), eq(CategoryEntityTable.isSystemCategory, false));
+
+        if (trimmed === '') {
```
liaugust: isEmptyString

packages/contracts/src/@generic/util/untranslated-where.util.ts
liaugust, line 1: extract into separate util files

packages/ai/src/embedding/service/embedding-suggestion.service.ts
```diff
@@ -165,9 +208,37 @@ export class EmbeddingSuggestionService {
...
             .sort((first, second) => second.score - first.score);
+
+        return sorted
+            .slice(0, EMBEDDING_CATEGORY_SUGGESTION_LIMIT)
+            .map(row => categories.find(category => category.id === row.categoryId))
+            .filter(isDefined);
+    }
+
+    private blendMccScores(scoreMap: Map<number, number>, mccRows: { categoryId: number; count: number }[]): void {
+        if (mccRows.length === 0) {
```
liaugust: isEmptyArray

packages/app/src/ai/enum/ai-system-action.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/enum/ai-system-state.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/enum/ai-system-umbrella-state.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/enum/drainer-kind.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/enum/drainer-state.enum.ts
liaugust, line 1: Why not upercase everything?

packages/app/src/ai/interface/chat-snapshot.interface.ts
liaugust, line 1: why do we need to re-export anything like this?
I think we can remove this file

packages/app/src/ai/service/ai-coordinator.service.ts
```diff
@@ -0,0 +1,143 @@
...
+            }
+            if (this.snapshot.isSuspended) {
+                this.setSnapshot({ isSuspended: false });
+                void this.startSubsystems();
+            }
+
+            return;
+        }
+
+        if (this.releaseTimer !== null) {
```
liaugust: isDefined

packages/app/src/ai/service/ai-system-status.service.ts
```diff
@@ -0,0 +1,364 @@
...
+            translationDrainerService.subscribe(this.scheduleRecompute),
+            embeddingDrainerService.subscribe(this.scheduleRecompute),
+            embeddingProgressStore.subscribe(this.scheduleRecompute),
+            translationProgressStore.subscribe(this.scheduleRecompute)
+        ];
+    }
+
+    protected emptySnapshot(): AiSystemSnapshotInterface {
+        return EMPTY_SNAPSHOT;
+    }
```
liaugust, line 147: You are returning the same object (reference to same object)

liaugust (changes requested):

packages/app/src/transaction/interface/recurring-calendar-accumulator.interface.ts
```diff
@@ -0,0 +1,11 @@
+import type { RecurringCalendarEntryInterface } from './recurring-calendar-entry.interface';
+
+export interface RecurringCalendarAccumulatorInterface {
+    readonly entriesByDay: Map<number, RecurringCalendarEntryInterface[]>;
+    readonly forecastedEntriesByDay: Map<number, RecurringCalendarEntryInterface[]>;
+    readonly isCurrentMonth: boolean;
+    readonly today: number;
+    readonly daysInMonth: number;
+    totalAmount: number;
+    forecastedTotalAmount: number;
```
liaugust: why something is readonly and something is not?

vitalyiegorov:
@liaugust FYI — I just got handed a fresh CODEX review that lists 6 items (2× P0, 3× P1, 1× P2). Quick cross-check against current HEAD `a298b3a79`: all six are already addressed by the session-1 fixes pushed earlier today. Summary table (with current-state evidence):

| # | CODEX item | Current HEAD file + fix |
|---|---|---|
| 1 | P0 — Voice/STT eagerly initialized on tab mount | `(tabs)/_layout.tsx:80` mount gated on `isAiAvailable && isVoiceInputOpen` — overlay subtree (the source of `useStartStopStt` → `sttService.start()`) only mounts on user intent. Commits `61505520e`, `6f82b9ecf`. |
| 2 | P0 — Suspended AI reported as Healthy/Ready | Both derive waterfalls now branch explicitly on `SUSPENDED`/`IDLE` before the Ready/Healthy fallthrough (`ai-umbrella-status.service.ts:99`, `ai-system-status.service.ts:209`). Enum variants added; exhaustive `Record<Enum, T>` maps updated; `de/es/fr/uk` translations committed. Commits `35ea650bf`, `1b94ea45d`. |
| 3 | P1 — Imported updates don't restamp `needsEmbedding` | New `deriveEmbeddingFlag` in `@budgie/contracts` runs inside `transactionRepository.updateById` (`transaction.repository.ts:59`). Any update touching title/comment/type/fromAccountId/toAccountId restamps; any transition to `TRANSFER`/`ADJUSTMENT` clears. Commit `19f14215e`. |
| 4 | P1 — Repository cleanup queries not awaited | expo-sqlite's driver is sync (`BaseSQLiteDatabase<'sync', …>`) — `runner.run()` returns void, not a thenable; adding `await` fires `@typescript-eslint/await-thenable`. Fix instead: `clearResidueOnce` wraps the trio in `transactionAsync(db, async tx => { … })` so the caller's `await` serializes the full atomic cleanup (`embedding-drainer.service.ts:116`). Logs fire AFTER the transaction commits. Commit `19f14215e`. |
| 5 | P1 — `TransactionEntitySchema` lies about `needsEmbedding` | Select schema no longer marks it `.optional()` — faithful to the `NOT NULL DEFAULT false` column. Optionality moved to the create schema via `.extend({ needsEmbedding: boolean().optional() })`. Commits `eb742a11d`, `d7d82d85c`. |
| 6 | P2 — Convert-to-transfer doesn't reconcile embedding | Resolved transitively through #3 — `convertToTransfer` passes `type: TRANSFER/DEBT` to `updateById`, which runs `deriveEmbeddingFlag`: `TRANSFER`/`ADJUSTMENT` clears the flag, `DEBT` with semantic fields re-stamps. No belt-and-suspenders caller-side fix added on purpose — invariant lives in the repository. Commit `19f14215e`. |

If the CODEX output above was run against an older SHA, a fresh run at `a298b3a79` should produce no overlap with the six items. Full verification report (with exact current-HEAD code snippets per item) committed locally at `docs/plans/2026-04-20-codex-review-verification.md` if you want me to paste it inline — it's gitignored so not in the PR diff.

Plus the 14 round-2 comment fixes (type guards, UPPER_CASE enum migrations, CTE extraction, `emptySnapshot` spread, accumulator class, etc.) are all landed on `ui-fixes` and CI is re-running after the latest tsx/knip cleanup.

packages/app/src/ai/service/base-subsystem.service.ts
liaugust, line 1: split classes by separate files

liaugust (changes requested):

liaugust (approved):

PR #386, opened 2026-04-26
fix(app): erste pdf positional parser + dedup-on-edit

vitalyiegorov:
## Summary

Rebuilds Erste PDF import on **positional PDFKit data** (per-line `{text, x, y, page}` items) instead of fragile text-stream regex heuristics. Verified end-to-end on iPhone 16 Pro against all 12 real Erste statements; transaction counts match expectations within tolerance, balances reconcile cleanly.

## Why

The previous text-stream parser (`ErsteModernTextParser`) used regex to detect block boundaries (`POS`, `E-COMM`, `Polizze Nr.`, numeric prefixes, etc.) inside flattened text from `PDFPage.string`. Two structural problems:

1. **iOS-simulator and macOS PDFKit produce different `selectionsByLine()` output for the same PDF.** Same Apple framework, different Y-bucketing across platforms. macOS-tuned heuristics drifted off-by-one on device imports → `Grouped Erste layout mismatch: N content blocks for M dated rows` errors.
2. **Pattern enumeration was open-ended.** Every new statement variant (yearly with 21 pages, single-transaction Bareinzahlung, AREALIS-style numeric prefixes, two-account SEPA with overlapping vendors) needed another regex tweak.

The structural information was already in the PDF — we were throwing it away when extracting to flat text.

### Native module — `expo-pdf-text-extract` (Swift patch)
- New `extractTextItems(filePath)` AsyncFunction returning `[{text, x, y, width, height, page}]` per `PDFPage.selectionsByLine()`.
- Existing `extractText` retained, layout-aware via Y-bucket + X-sort + 12pt column-gap separator.
- `podspec` excludes `**/*Tests.swift` so XCTest-dependent files don't leak into the host app build.

### bank-sync — `ErsteModernPositionalParser`
- Reads positional items directly. Algorithm: filter footer (Y < 50) → sort by page asc, Y desc → bucket items into rows by Y tolerance → split each row into LEFT (X < 250) and RIGHT (X ≥ 250) columns → identify transaction anchors via `DD.MM.YYYY <amount>[-]?` regex on RIGHT column (or inline tail of full-width LEFT row for `Bareinzahlung`-style entries) → pair each anchor's LEFT items as primary line; subsequent LEFT-only rows until next anchor as continuation.
- Transaction state held in `TransactionAccumulator` and `ParserState` classes (rule 29: mutable accumulators are classes, not interfaces).
- Account info (IBAN, balances, statement date) extracted by direct positional lookup in `extract-erste-account-info.util`.
- Card-merchant city/country enrichment preserved via `parseErsteCardMerchant` running over continuation lines.
- Layout thresholds (Y row tolerance, right-column X, footer Y) live in `erste.constant.ts` as named constants — single source of truth.

### Layer-1 dedup-on-edit fix
- `transactionRepository.updateById` now accepts a narrowed `TransactionUpdateInputInterface` that omits `externalId` / `externalSource`. Regular form edits (tags, category, comment, amount) cannot accidentally clear the bank-sync linkage, so reimport hits the existing row by `externalId` and updates in place.
- Call site (`transactionService`) field-picks the allowed properties before passing to the repo.

### App
- `packages/app/src/sync/util/extract-pdf-text-items.util.ts` thin JS wrapper around the native API.
- `erste-sync.service.ts` calls `extractTextItems` then delegates to `ErsteFileClient.parse(items)`. `@Log` decorator on `parseFile` only — ad-hoc diagnostic logs and dump scaffolding removed.
- `base-file-bank-sync.service.ts` member ordering fixed (public decorated → private decorated → private non-decorated).

### Cleanup (deletions)
- `erste-modern-text.parser.ts` + base + interface — replaced by positional parser.
- `parse-erste-pdf.util.ts`, `parse-erste-statement-date.util.ts`, `parse-erste-modern-date-amount.util.ts` — replaced by `parse-erste-items.util` and `extract-erste-account-info.util`.
- `erste-modern-*-transaction-state.interface.ts` (×2), `erste-modern-date-amount-input.interface.ts` — text-parser-only.
- `ERSTE_IBAN_REGEX`, `ERSTE_ACCOUNT_NUMBER_REGEX`, `ERSTE_MODERN_FORMAT_MARKER`, `ERSTE_MODERN_TRANSACTION_DATE_REGEX`, `ERSTE_MODERN_INLINE_TRANSACTION_TAIL_REGEX`, `ERSTE_MODERN_FULL_DATE_REGEX`, `ERSTE_MODERN_END_MARKER`, `ERSTE_MODERN_BALANCE_*` — text-parser-only.
- `packages/bank-sync/jest.config.js`, `test`/`test:watch`/`test:cov` scripts — per CLAUDE.md rule 27 (no unit tests).

## Verified

- All 12 statements imported on iPhone 16 Pro: 2024001=511, 2025001=81, 2026001=145, 2026002=42, 2026003=3, 2026004=1, 2026005=2, 2026006=1, 2026007=10, 2026008=11, 2026009=5, 2026010=8 transactions — within expected ranges.
- Balance + Net Worth reconcile via SQL: `account_balances` snapshot = sum of 823 entries = €11,364.25; net worth = €11,364.25 × 1.17096 (EUR→USD) = $13,307.08, displayed correctly.
- `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd` — green.

## Migration note

External-id hash includes `details` (carried over from the SEPA-collision fix). On first reimport against an existing pre-PR database, rows with the old hash won't match by `externalId` and may appear as new transactions. Either accept the one-time duplication or wipe + re-import.


packages/bank-sync/src/erste/parser/erste.parser.ts
liaugust, line 1: fix formating

liaugust, line 1: why not to move each util to private methods inside class ?

packages/bank-sync/src/erste/parser/transaction-accumulator.ts
```diff
@@ -0,0 +1,51 @@
...
+    for (const line of lines) {
+        const merchant = parseErsteCardMerchant(line);
+
+        if (isDefined(merchant)) {
+            return { city: merchant.city, countryAlpha2: merchant.countryAlpha2 };
+        }
+    }
+
+    return {};
+};
```
liaugust: let's make it private method inside TransactionAccumulator

packages/bank-sync/src/erste/util/extract-erste-account-info.util.ts
liaugust, line 1: let's split by separate files

packages/bank-sync/src/erste/util/group-pdf-items-into-rows.util.ts
liaugust, line 1: let's split by separate files

liaugust (changes requested):

liaugust (approved):

PR #408, opened 2026-05-03
docs: refund-pair consolidation design (#243)

### LINKED ISSUE #243: Refund transactions
### 📌 Problem
Users cannot properly handle refunds today, which leads to incorrect balances and misleading spending insights when purchases are refunded.

---

### 🎯 Goal
Allow users to represent refunds in their transaction history in a clear and intuitive way.

---

### 🧩 User Scenarios
- [ ] **Mark existing transaction as refund**  
  Reclassify a transaction when the original purchase is no longer relevant.

- [ ] **Create refund from existing transaction**  
  Create a new refund entry linked to the original transaction, while keeping the original intact.

- [ ] **Unlink refund from existing transaction**  
  Unlink a refund entry from the original transaction.

---

### 🖥️ User Experience
- [ ] Provide actions: **Mark as refund** / **Create refund**
- [ ] Explain the difference between both options
- [ ] Visually distinguish refund transactions in lists
- [ ] Indicate relationship between refund and original transaction

---

### 📊 App Behavior
- [ ] Refunds correctly offset balances
- [ ] Refunds do not inflate expenses or income
- [ ] Spending insights and summaries remain accurate
- [ ] Works fully offline

---

### 🔮 Future Improvements (Optional)
- [ ] Show refunded vs remaining amount
- [ ] Group refunds with original transaction
- [ ] Suggest potential refunds automatically
- [ ] Allow more than one refund for the same transaction
- [ ] Allow refunding only part of the original amount

## Creating refund transaction for original transaction

Image

Image

Image

Image

Image

## Marking existing transaction as a refund for another

Image

Image

Image

Image

vitalyiegorov:
## Summary

- Design spec only (no code yet) for **auto-detecting bank-synced refund pairs** during the existing consolidation pass.
- One expense + N later refund incomes on the **same account** with **same currency** consolidate into a synthetic net-expense canonical (`type=EXPENSE`, `consolidationType=REFUND`); sources reparent + hide via `consolidationParentTransactionId`, identical mechanism to today's `TRANSFER_PAIR` / `IBAN_BRIDGE_TRANSFER` / `ATM_CASH_WITHDRAWAL`.
- Adds two confidence buckets: **auto** (strict `UPPER(TRIM(title))` match, ≤ 30 days) and **manual review** (title containment after stripping `REFUND` / `RETURN` / `REVERSAL` / `CHARGEBACK` / `CR ` prefixes, matching MCC, ≤ 90 days). Manual review reuses the existing `consolidation-source.tsx` modal.
- **No schema migration.** Adds `TransactionConsolidationTypeEnum.REFUND`, new candidate interfaces, two time-window constants, a prefix list, and a sibling `transaction-refund.repository.ts`. All data rides on existing columns.

## Why heuristic only

Researched the Monobank Personal API while drafting: `StatementItem` exposes **no** explicit refund / reversal / chargeback signal. Every documented field is already in `BaseTransactionFieldsInterface`. Title + amount + time + MCC + same-account heuristics are the only viable detection path. Documented in the spec.

## Explicit deferrals

Locked into the spec so the next PR doesn't drift:

- **Manual UX flows from #243** (Mark as refund / Create refund / Unlink). Data model here doesn't block them.
- **Cross-currency refunds.**
- **Cross-account refunds** (refund on a replacement card).
- **Late-arriving refunds** (synced after the expense is already consolidated).

## Test plan

- [ ] Spec reviewed against `transferConsolidationService` shape — confirmed it slots in as a 4th processor in `runConsolidation` without touching the existing three.
- [ ] Verified `use-get-statistics-transactions.query.ts` filters positive amounts so a 0-amount full-refund canonical stays hidden from spending insights.
- [ ] 9 integration scenarios planned under `tests/bank-sync-tests/src/scenarios/consolidation/` covering: strict match, partial, 1:N, full refund, time-window boundaries, manual review fallback, transfer-pair exclusion, cross-currency exclusion, unconsolidate symmetry.
- [ ] No new ESLint disables expected; SQL CTE file may need an approved `max-lines` disable per root rule 4 if it grows.

Closes #243

### DIFF
--- a/docs/superpowers/specs/2026-05-03-refund-consolidation-design.md
+++ b/docs/superpowers/specs/2026-05-03-refund-consolidation-design.md
@@ -1,0 +1,386 @@
+# Refund-pair consolidation — design
+
+**Branch:** `feat/refund-consolidation`
+**Issue:** [#243 — Refund transactions](https://github.com/budgie-at/budgie/issues/243)
+**Scope:** Auto-detect bank-synced refund pairs (one prior expense + one or more later refund incomes on the same account) during the existing consolidation pass, and merge them into a synthetic net-expense canonical that hides the source rows. Manual UX flows from the issue (Mark as refund / Create refund / Unlink) are deferred.
+
+## Goal
+
+Refunds today inflate income, leave the original expense at full magnitude, and confuse spending insights even though card balances are mathematically correct. The user-visible problem: spending charts show the full €120 purchase plus €120 income, instead of showing the €0 effective spend. The issue lists three manual flows; we are intentionally **not** building them in v1. Instead we extend the existing consolidation engine to recognise refund pairs the same way it already recognises transfer pairs, IBAN-bridge transfers, and ATM withdrawals, and produce a single net-expense canonical that the rest of the app already treats as the source of truth (sources reparented, hidden by filtered indexes, balances correct).
+
+## Non-goals
+
+- **Manual UX flows** (`Mark as refund`, `Create refund from expense`, `Unlink`) — deferred. The data model we land here does not block them.
+- **Visible "linked" refund pair** — the issue's "Group refunds with original transaction" / "Visually distinguish refund transactions" copy applies to a model that keeps both rows visible. We chose the canonical-hidden-source model instead, matching every other consolidation type today. Both rows are reparented; only the canonical is visible.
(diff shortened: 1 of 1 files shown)


vitalyiegorov:
Closing — specs in docs/superpowers/specs/ are gitignored by project convention; force-adding them was a mistake. Design work continues locally; implementation PR will land separately on feat/refund-consolidation-impl off latest main.

Issue #444, opened 2026-05-17
Sync-time UI lag: phase 2 follow-ups (bulk update, progress throttle, live-query debounce)


liaugust:
Phase 2 of the sync-time UI lag work. Phase 1 (microPause yields in `prepareCreateInputsForRules` + `updateExistingTransactions`) shipped in #437 / commit `ca5655653`. This issue tracks the three remaining items deferred from #440.

## Suggested ordering
1. **F2 first** — smallest, highest UX-impact (cuts re-render storm at source).
2. **F1 second** — medium effort, eliminates per-row native-bridge cost.
3. **F3 last** — only needed if F2 doesn't reduce the storm enough; might end up scoped down or skipped.

This ordering lets us measure after F2 alone (with phase-1 microPause yields in place) and reassess whether F1 and F3 are still needed.

---

## F1 — Bulk UPDATE in `updateExistingTransactions`

**File**: `packages/app/src/sync/service/monobank-sync.service.ts:333-346`

**Change**: replace the `for...of await transactionService.update(...)` loop with a single repository call that issues one SQL `UPDATE ... CASE WHEN ... END WHERE id IN (...)` (or batches via `INSERT ... ON CONFLICT DO UPDATE`).
- Add `transactionService.bulkUpdate(inputs)` and underlying `transactionRepository.bulkUpdate(...)` accepting `TransactionCreateInputInterface[]`.
- Wrap in one `transactionAsync` so the existing connection pattern is preserved.
- Drop the per-iteration `microPause()` added in phase 1 since the loop becomes a single statement.

**Acceptance**
- A sync page with N existing transactions issues 1 `UPDATE` (not N).
- Updated rows: `amount`, `exchangeRate`, `toIban` (the only fields the current `update()` writes).
- `tests/bank-sync-tests/src/scenarios/monobank/unchanged-data-noop.test.ts` continues to pass.
- New test: Monobank fixture with 5 changed amounts on existing transactions verifies all 5 are updated in 1 SQL statement.

**Risk**: existing `transactionService.update` is used elsewhere. Don't touch the original; add a new bulk method, only the sync orchestrator switches.

**Complexity**: medium (~80–120 lines: repo method + service method + sync swap + test).

---

## F2 — Throttle `bank_syncs` progress writes

**File**: `packages/app/src/sync/service/monobank-sync.service.ts` (search for mid-batch `bankSyncRepository.update` callsites that write `forwardSyncedAt` / `transactionCount` / `status`).

[...]

PR #460, opened 2026-05-21
fix(app): serialize SQLite close + sequentialize rule batch writes

### LINKED ISSUE #459: App crash: concurrent SQLiteModule.closeDatabase race during rekey/import (SEV-0)
## Summary

iOS/TestFlight users are crashing during destructive DB operations (rekey, import) with multiple threads inside `SQLiteModule.closeDatabase` simultaneously. Four crash logs collected from 5.17.1 (40), all matching the same signature.

## Crash signature

```
Process:     budgie
Version:     5.17.1 (40)
OS:          iPhone OS 26.4.2 (23E261)
Crash sites: SQLiteModule.closeDatabase  →  sqlite3Close  →  sqlite3SchemaClear  →  deleteTable
             (multiple threads, simultaneously)
```

Failure modes observed across the 4 crashes:
- `EXC_BAD_ACCESS (SIGBUS) / KERN_PROTECTION_FAILURE` — raw use-after-free on torn-down page.
- `EXC_BREAKPOINT (SIGTRAP)` — Swift ARM pointer-authentication trap in `swift_unknownObjectRetain` catching corrupted object metadata.
- `EXC_BAD_ACCESS (SIGSEGV) / KERN_INVALID_ADDRESS` — dangling pointer in a 48GB unmapped VM region.

All four traces show 3–5 threads simultaneously inside `SQLiteModule.closeDatabase → Collection.firstIndex(where:)` at `SQLiteModule.swift:492`. The crashing thread iterates the module's open-DB array while another thread mutates it.

## Root cause

Two open native SQLite handles to the same DB file:
1. `packages/app/src/@generic/drizzle/db/db.ts:45` — `global.__expoSqliteDb__` opened via `SQLite.openDatabaseSync(DB_NAME, ...)`.
2. `packages/app/src/app/root-layout-content.tsx:78` — `<SQLiteProvider databaseName={DB_NAME}>` opens a second handle (used by `useSQLiteContext()` at `(tabs)/index.tsx:42`).

Close paths that fire concurrently:
- `database-rekey.service.ts:43` — PIN-change rekey
- `database-import.service.ts:36` — backup restore
- SQLiteProvider unmount on `reloadApp()` after import

When two of these overlap (rekey/import + provider unmount on reload), the native module sees concurrent `closeAsync` calls on overlapping handles. The use-after-free in `Collection.firstIndex(where:)` is the crash site.

## Fix

1. Drop `<SQLiteProvider>` from `root-layout-content.tsx`. Migrate the lone `useSQLiteContext()` consumer at `(tabs)/index.tsx:42` to `import { expoDb }` directly. Removes the dual-handle footgun.
2. New `databaseLifecycleService.close()` with a module-level single-flight `Promise<void>` lock. Multiple callers reuse the in-flight promise; the native module never sees concurrent close calls.
3. Pre-close drain: await `monobankSyncService.whenIdle()` and `ruleApplicationDrainerService.whenIdle()` before closing. Avoids closing the DB while a write transaction is in flight.
4. Migrate `database-rekey.service.ts` and `database-import.service.ts` to use `databaseLifecycleService.close()`.

## Verification

- Manual: trigger rekey while sync is running; trigger import; rapid backgrounding during a destructive operation. No crash on the patched build.
- TestFlight: monitor crash reports for `SQLiteModule.swift:492` signatures after rollout.

## Related

- Issue #444 (sync UI lag phase 2) — separate problem; sync-time UI lag is addressed by sequential rule batch + balance-update coalescing.

liaugust:
## Summary

- Fixes #459 (SEV-0): concurrent `SQLiteModule.closeDatabase` race causing iOS crashes during rekey/import — four crash reports collected on 5.17.1 (40), all matching the same signature.
- Partially addresses #444: sequentializes rule-engine per-batch writes (removes 20-deep WAL writer storm) and hoists `updateAllBalances` out of the per-row transaction.

## Scope clarification

This PR ships **CS0 (lifecycle race) and CS1 (rule-engine concurrency)** from the consensus plan.

The remaining sync-UX work — `SyncBalanceUpdateMode` enum across 17 callsites, `bank_syncs` progress throttle, bulk UPDATE in `updateExistingTransactions`, and the `useDebouncedLiveQuery` wrapper (#444 items F1–F3) — is deferred to a follow-up PR so we can measure post-CS0/CS1 profiler baselines before deciding whether F3 is still needed.

### CS0 — DB lifecycle single-flight (fixes #459)

- **Drop `<SQLiteProvider>` from `root-layout-content.tsx`** — the second native handle is what made the close race possible. The lone `useSQLiteContext()` consumer at `(tabs)/index.tsx:42` migrates to a direct `import { expoDb }`.
- **New `databaseLifecycleService.close()`** in `packages/app/src/@generic/drizzle/service/database-lifecycle.service.ts` — module-level `closeInflight: Promise<void> | null` lock. Multiple callers reuse the in-flight promise; the native module never sees concurrent close calls.
- **Pre-close drain** via new `monobankSyncService.whenIdle()` and `ruleApplicationDrainerService.whenIdle()`. Avoids closing the DB while a write transaction is in flight.
- **Migrate close call sites**: `database-rekey.service.ts:commit()` and `database-import.service.ts:replaceFromUri()` now call `databaseLifecycleService.close()` instead of `expoDb.closeAsync()` directly. The `clearDatabaseGlobals()` helpers are deleted (now centralized in the lifecycle service).

### CS1 — Rule-engine sequential per-row writes (partially addresses #444)

- **`applyRuleToMatchingTransactionBatch`** — replaces `Promise.allSettled(batchIds.map(...))` with a sequential `for...of await`. Per-row `transactionAsync` is preserved so partial-failure semantics are intact (1 bad row → 19 still apply, `failed = 1`).
- **`applyRulesToTransactionsBatch`** — replaces `Promise.all(batchIds.map(...))` with the same sequential pattern.
- **Hoisted balance update**: `accountBalanceIncrementalService.updateAllBalances(true, transaction)` is no longer called inside each per-row `transactionAsync`. The inner methods (`applyRulesToTransaction`, `applyRuleActionsToTransaction`) now return a boolean indicating whether the row converted to a transfer; the outer orchestrators aggregate across batches and call `updateAllBalances(true)` once if any row converted.
- New `RuleBatchOutcomeInterface` (`{ failed: number; converted: boolean }`) for batch-result aggregation.

## Verification

- `yarn ts` — green
- `yarn lint` — green (1 pre-existing warning in `use-suggestion-base.hook.ts` unrelated to this PR)
- `yarn deadcode` — green
- `yarn cpd` — green
- `yarn format` — applied
- `tests/bank-sync-tests` — 24 files / 42 tests pass

## Test plan

- [ ] **Manual SEV-0 repro on `main`**: trigger a rekey (PIN change) while a sync is mid-flight, or trigger an import then rapidly background the app. Capture native crash log with the `SQLiteModule.swift:492` signature.
- [ ] **Same repro on this branch**: verify no crash, single sequential close, app reload completes cleanly.
- [ ] **Rule >50 txns**: create a `Title CONTAINS X` rule matching 100+ existing transactions, apply via the rules UI, verify no crash, all 100 categorized, balances correct.
- [ ] **Rule partial-failure semantics preserved**: with a 100-txn fixture containing one bad row (e.g., a missing entry), verify 99 categorized + 1 reported failed via the existing `applyRuleResult.failed`.
- [ ] **Lifecycle stress**: rekey while sync is running → verify drain completes before close; backgrounding during import → verify single-flight lock collapses concurrent close calls.
- [ ] **Sync still works**: clean install → seed monobank account → run first sync → verify transactions appear, balances correct.


packages/app/src/rule/service/rule-application-drainer.service.ts
```diff
@@ -84,6 +84,13 @@ class RuleApplicationDrainerService {
         }
     }
 
+    async whenIdle(): Promise<void> {
```
vitalyiegorov, line 87: we already have run/ drainPending, why we need additional method? how awaiting it here will help?

packages/app/src/rule/service/rule-engine.service.ts
```diff
@@ -56,11 +57,16 @@ class RuleEngineService {
...
-        await this.getBatchStarts(transactionIds.length).reduce(
-            (previousBatch, batchStart) =>
-                previousBatch.then(() => this.applyRulesToTransactionsBatch(batchStart, transactionIds, evaluationInputs, rules)),
-            Promise.resolve()
-        );
+        let converted: boolean = false;
+        for (const batchStart of this.getBatchStarts(transactionIds.length)) {
+            // eslint-disable-next-line no-await-in-loop -- sequential per-batch processing to avoid concurrent SQLite writers
+            const batchConverted = await this.applyRulesToTransactionsBatch(batchStart, transactionIds, evaluationInputs, rules);
+            converted ||= batchConverted;
```
vitalyiegorov, line 64: what is this?

packages/app/src/rule/service/rule-engine.service.ts
```diff
@@ -56,11 +57,16 @@ class RuleEngineService {
...
+        }
+
+        if (converted) {
+            await accountBalanceIncrementalService.updateAllBalances(true);
```
vitalyiegorov, line 68: how this helps with crashes? this can only create more?

packages/app/src/rule/service/rule-engine.service.ts
```diff
@@ -113,45 +120,55 @@ class RuleEngineService {
         }
 
         const total = matchingIds.length;
-        const failed = await this.applyRuleToMatchingTransactionBatches(matchingIds, rule.actions, onProgress);
+        const batchOutcome = await this.applyRuleToMatchingTransactionBatches(matchingIds, rule.actions, onProgress);
 
-        const applied = total - failed;
+        if (batchOutcome.converted) {
+            await accountBalanceIncrementalService.updateAllBalances(true);
```
vitalyiegorov, line 126: why do you think adding FULL balace recalculation can help with crash or hangs?

packages/app/src/sync/service/monobank-sync.service.ts
```diff
@@ -368,6 +370,16 @@ class AppMonobankSyncService {
             forwardSyncedAt: null
         });
     }
+
+    private async runSync(): Promise<BackgroundTask.BackgroundTaskResult> {
```
vitalyiegorov, line 374: this method is obsolete, why do we need it?  this whole fixes look super strange and over engineered.

vitalyiegorov (changes requested):

packages/app/src/@generic/drizzle/service/database-lifecycle.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    private async runClose(): Promise<void> {
+        await this.drainInFlightWork();
+
+        await expoDb.closeAsync();
+        global.__expoSqliteDb__ = undefined; // eslint-disable-line no-undefined, no-underscore-dangle
+        global.__drizzleDb__ = undefined; // eslint-disable-line no-undefined, no-underscore-dangle
+    }
+
+    private async drainInFlightWork(): Promise<void> {
+        await Promise.all([monobankSyncService.whenIdle(), ruleApplicationDrainerService.whenIdle()]);
```
vitalyiegorov, line 36: This is not a complete DB quiesce barrier yet. It waits for the current Monobank sync promise and the rule drainer's current `runPromise`, but it does not cover scheduled rule work, pending rule queues, the rule idle callback/timer, transfer-consolidation work enqueued by sync, or the AI drainers that also write to the DB. That means `close()` can still replace/close the database while queued DB writers run immediately afterward. For a SEV-0 close race fix, this needs to either drain/cancel every DB writer source or stay narrowly scoped to single-flight close plus removing the second SQLite handle.

packages/app/src/rule/service/rule-application-drainer.service.ts
vitalyiegorov, line 87: This method only waits while `runPromise` is defined. It does not start `drainPending()`, does not cancel the 250ms timer, does not wait for `cancelIdleCallback`, and does not inspect `pendingTransactionIds` / `pendingRuleIds`. So awaiting it here can return even though rule work is queued and about to run. That makes the lifecycle drain look stronger than it is.

packages/app/src/rule/service/rule-engine.service.ts
vitalyiegorov, line 68: This hoist changes failure semantics. If an earlier row converts to transfer and a later row in `applyRulesToTransactionsBatch` throws, the earlier row has already committed but this outer `updateAllBalances(true)` is skipped because the method rejects. The drainer catches the rejection and continues, leaving balances stale. Previously the balance update happened inside the successful row transaction, so successfully converted rows also recalculated balances even when a later row failed.

packages/app/src/rule/service/rule-engine.service.ts
vitalyiegorov, line 126: This full balance recalculation is not part of the SQLite close crash fix, and it adds more DB work to a PR whose stated goal is reducing lifecycle/write pressure. If this change is kept for the rule-performance slice, it needs a clear correctness argument and failure-safety story. Otherwise it should stay out of the lifecycle PR.

packages/app/src/@generic/drizzle/service/database-rekey.service.ts
```diff
@@ -40,8 +42,7 @@ class DatabaseRekeyService {
     }
 
     private async commit(paths: RekeyPathsInterface, onCommit: () => Promise<void>): Promise<void> {
-        await expoDb.closeAsync();
-        this.clearDatabaseGlobals();
+        await databaseLifecycleService.close();
```
vitalyiegorov, line 45: Import pauses translation/embedding drainers before closing, but rekey now only goes through `databaseLifecycleService.close()`, whose drain does not pause those AI drainers. If those services can write during import, they can also write during rekey. Centralizing close without centralizing the same runtime pause makes the lifecycle contract inconsistent.

packages/app/src/sync/service/monobank-sync.service.ts
```diff
@@ -222,6 +217,13 @@ class AppMonobankSyncService {
         return result;
     }
 
+    async whenIdle(): Promise<void> {
```
vitalyiegorov, line 220: Waiting for the whole sync is a coarse proxy for waiting on DB writes. It includes network calls, recursive sync-loop work, and rate-limit pauses, so destructive operations can appear hung even when no transaction is active. If the goal is to avoid close-during-write, the lifecycle contract should track active DB write sections or explicitly pause/cancel the sync scheduler, not await an entire background sync loop.

vitalyiegorov (changes requested):
I agree with the existing review direction. The implementation removes the extra SQLiteProvider handle, but the new lifecycle and rule changes do not yet make the destructive DB operations safe. The main gaps are: the pre-close drain does not cover scheduled/pending writers, transfer consolidation is not drained at all, rekey does not pause the AI drainers that import already pauses, and the balance-update hoist changes partial-failure semantics. I would simplify this toward the minimal crash fix: remove the second handle, single-flight close, and either fully quiesce/cancel all DB writers before close or leave unrelated rule-engine/balance behavior out of this PR.

vitalyiegorov:
Suggested robust direction for the fix:

This should be modeled as an explicit DB lifecycle gate, not as scattered `whenIdle()` waits. Destructive DB operations need one owner that can move the app into a `closing/replacing` state, reject or defer new DB work, cancel scheduled drainers/timers/idle callbacks, wait for active DB write sections to finish, then close exactly one SQLite handle.

Concretely:

1. Keep the minimal handle fix: remove `<SQLiteProvider>` and use the shared `expoDb` everywhere.
2. Add a small DB lifecycle/write gate used by DB-writing services. It should track active write sections and expose something like `runWrite(...)` / `pauseForClose(...)`, or an equivalent centralized protocol.
3. Destructive operations should first block new writes, then cancel scheduled work queues, then drain active writes, then close single-flight.
4. The services that currently schedule delayed DB work need explicit lifecycle methods, not passive `whenIdle()` loops: rule application drainer, transfer consolidation drainer, Monobank sync, embedding drainer, translation drainer.
5. Keep rule-engine balance recalculation changes out of the SEV-0 close-race PR unless they are required for the crash fix. If kept separately, they need failure-safe semantics: a committed conversion must not leave balances stale if a later row fails.

The current PR has the right first move by removing the second SQLite handle, but the rest should converge on one explicit quiesce/close contract. Otherwise we will keep adding partial drains that look safe locally while missing scheduled or newly enqueued writers.

liaugust:
Splitting into three focused PRs: rules sequentialization, SQLiteProvider removal, and SQLite close single-flight. New PRs incoming.

PR #466, opened 2026-05-21
fix(app): consolidate legacy transfer leftovers

vitalyiegorov:
## Summary
- Consolidate legacy same-bank hinted fee transfers when imported accounts only have IBAN bank hints and no external source.
- Attach orphan bridge source/target rows to an existing IBAN bridge canonical transfer instead of leaving technical duplicates visible.
- Auto-consolidate first-occurrence interbank fee transfers when the match is uniquely determined by transfer MCC, different bank sources, same currency, close time, and bounded fee delta.
- Add bank-sync regression coverage for same-bank fee transfers, bridge canonical duplicate leftovers, and interbank fee transfers.

## Root Cause
Some historical imports leave raw technical transfer rows beside an already-created canonical bridge transfer. The generic bridge matcher only creates or reuses a direct transfer while processing bridge rows; it did not have an attach-only path for orphan source and target rows that already match a canonical `IBAN_BRIDGE_TRANSFER` / `IBAN_BRIDGE_CHAIN_TRANSFER`.

Interbank card transfers can also arrive as a charged source-side expense and a net target-side income from different banks. There is no user-visible manual path to consolidate these once imported, so the matcher needs a conservative auto path for first occurrences instead of relying on prior canonical history.

## Validation
- `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/consolidation`
- `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd`

Note: lint still reports the pre-existing warning in `packages/app/src/ai/hook/use-suggestion-base.hook.ts:101`; the command exits successfully.


packages/contracts/src/transaction/repository/transfer-pair.repository.ts
liaugust, line 1: The file is huge already - consider splitting it into smaller ones

tests/bank-sync-tests/src/scenarios/consolidation/same-bank-hinted-fee-transfer.test.ts
liaugust, line 1: Let's extract test-utils into separate files

1. test files are growing super fast
2. utils are duplicated across the files
3. it is hard to understand the files when there is such a mess with different utils

It is not about exactly this file, but about our test files in general

liaugust (changes requested):

packages/contracts/src/transaction/repository/transfer-pair.repository.ts
vitalyiegorov: Addressed by splitting the query construction out of `TransferPairRepository` into focused SQL factory files. The repository now just logs and delegates each query to the DB client.

tests/bank-sync-tests/src/scenarios/consolidation/same-bank-hinted-fee-transfer.test.ts
vitalyiegorov: Addressed by extracting the shared consolidation test helpers into the bank-sync harness: `expectSingleConsolidation`, `seedBankSyncAccount`, and `updateBankTransaction`. The affected consolidation scenarios now reuse those helpers instead of local duplicates.

liaugust (approved):

PR #468, opened 2026-05-21
fix(app): drop SQLiteProvider to eliminate dual-handle close race

liaugust:
## Summary

Fixes a major contributor to the SQLite close-race crash (#459). The app currently opens **two native SQLite handles to the same DB file**:

1. \`packages/app/src/@generic/drizzle/db/db.ts:45\` — \`global.__expoSqliteDb__\` via \`SQLite.openDatabaseSync(DB_NAME, ...)\` (used by drizzle and every repository).
2. \`packages/app/src/app/root-layout-content.tsx:78\` — \`<SQLiteProvider databaseName={DB_NAME}>\` (used by \`useSQLiteContext()\` in one place).

When destructive operations (rekey, import) call \`closeAsync()\` on one handle while React unmounts the provider tree (e.g., during \`reloadApp()\` after import), **both handles' \`closeAsync()\` invocations dispatch onto separate GCD worker threads and race through the native module's open-DB array** — observed as concurrent threads in \`SQLiteModule.closeDatabase → Collection.firstIndex(where:)\` at \`SQLiteModule.swift:492\` in five TestFlight crash reports (SIGBUS, SIGTRAP/PAC, SIGSEGV).

This PR removes the provider entirely.

## What changed

- \`packages/app/src/app/root-layout-content.tsx\` — removed \`<SQLiteProvider>\` wrapping, removed unused \`DB_NAME\` import and \`SQLOptions\` const.
- \`packages/app/src/app/(tabs)/index.tsx\` — the sole \`useSQLiteContext()\` consumer (line 42, used only to hand the handle to \`useDrizzleStudio\`). Migrated to \`import { expoDb } from '.../db'\` + \`useDrizzleStudio(expoDb)\`.

## Why this is safe

- The provider's handle was never used by the data layer; drizzle and all 23 repository singletons already use \`global.__expoSqliteDb__\` from \`db.ts\`.
- The provider had exactly one consumer in the app (verified via grep), used in a way that doesn't depend on React context — \`useDrizzleStudio\` just needs the SQLite handle.

## Scope

This PR fixes one half of the close race (the unnecessary second handle). A follow-up PR will introduce a single-flight lock on the remaining close paths (rekey + import) as defense in depth.

Refs #459

## Test plan

- [ ] App launches normally.
- [ ] Home tab (\`(tabs)/index.tsx\`) renders; \`useDrizzleStudio\` still functions in dev (or no-op in prod as before).
- [ ] Import a backup → app reloads cleanly, no crash with the \`SQLiteModule.swift:492\` signature in the device console.
- [ ] PIN change (rekey) → app reloads cleanly, no crash.
- [ ] \`yarn ts && yarn lint && yarn deadcode && yarn cpd\` green.

### DIFF
--- a/packages/app/src/app/(tabs)/index.tsx
+++ b/packages/app/src/app/(tabs)/index.tsx
@@ -1,6 +1,5 @@
 import { AccountTypeEnum } from '@budgie/contracts';
 import { useDrizzleStudio } from 'expo-drizzle-studio-plugin';
-import { useSQLiteContext } from 'expo-sqlite';
 import { View } from 'react-native';
 import { useSharedValue } from 'react-native-reanimated';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
@@ -10,6 +9,7 @@ import { isNotEmptyArray } from '@rnw-community/shared';
 import { AnimatedSectionList } from '../../@generic/component/animated-section-list/animated-section-list';
 import { CollapsibleHeader } from '../../@generic/component/collapsible-header/collapsible-header';
 import { FLOATING_TAB_BAR_HEIGHT, FLOATING_TAB_BAR_MARGIN } from '../../@generic/constant/floating-tab-bar.constant';
+import { expoDb } from '../../@generic/drizzle/db/db';
 import { AccountGridItem } from '../../account/component/account-grid-item/account-grid-item';
 import { AccountSectionHeader } from '../../account/component/account-section-header/account-section-header';
 import { AccountsEmptyState } from '../../account/component/accounts-empty-state/accounts-empty-state';
@@ -39,8 +39,7 @@ export default function HomePage() {
     const { accounts } = useAccountsWithBankSyncQuery();
     const { bottom } = useSafeAreaInsets();
 
-    const db = useSQLiteContext();
-    useDrizzleStudio(db);
+    useDrizzleStudio(expoDb);
 
     const scrollY = useSharedValue(0);
 
--- a/packages/app/src/app/root-layout-content.tsx
+++ b/packages/app/src/app/root-layout-content.tsx
@@ -3,7 +3,6 @@ import { i18n } from '@lingui/core';
 import { useMigrations } from 'drizzle-orm/expo-sqlite/migrator';
 import { Stack } from 'expo-router';
 import * as SplashScreen from 'expo-splash-screen';
-import { SQLiteProvider } from 'expo-sqlite';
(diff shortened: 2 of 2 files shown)


vitalyiegorov (approved):

vitalyiegorov:
I would not treat this PR as proven crash fix yet. It is a reasonable cleanup, but it may also explain why import now crashes more reliably.

Hypothesis: the existing `<SQLiteProvider>` was accidentally acting as a second reference to the same native DB. In current `expo-sqlite`, same-path/same-options opens are cached/ref-counted unless `useNewConnection` is set. With the provider still mounted, `expoDb.closeAsync()` during import may only decrement the native ref count and avoid the real native `sqlite3_close`. After this PR removes the provider, `expoDb.closeAsync()` can perform the actual native close immediately.

That matters because our DB init loads `sqlite-vec` and creates `vec0` virtual tables. Expo has an open close-time crash report around SQLite virtual-table cleanup / statement finalization during `SQLiteModule.closeDatabase`:

- https://github.com/expo/expo/issues/38168

Expo also has history around same-DB multiple handles and close ref-counting:

- https://github.com/expo/expo/issues/33677
- https://github.com/expo/expo/pull/35818

So the observed behavior can be: this PR removes the extra handle, but by doing so it exposes the real close-path crash during import. That would make this cleanup directionally correct, but not sufficient and potentially unsafe to ship alone.

Concrete checks before merging:

1. Capture the native import crash stack on this branch. If it shows `SQLiteModule.closeDatabase`, `sqlite3Close`, `sqlite3VtabDisconnect`, `exsqlite3_finalize`, or other virtual-table cleanup frames, the root cause is likely close-time cleanup, not the provider itself.
2. Temporarily open our main DB with `finalizeUnusedStatementsBeforeClosing: false` in addition to `enableChangeListener: true`. If import stops crashing, it strongly points to Expo issue #38168.
3. Temporarily disable `sqlite-vec` extension loading / `vec0` virtual table creation and retry import. If crash disappears, that confirms the virtual-table close path.
4. Test the same import flow with the provider restored. If restored provider avoids the crash, that means the second ref was masking real close rather than causing this specific import crash.

My recommendation: keep this as optional cleanup, but do not ship it alone as the SEV-0 fix. The robust import/rekey fix should first establish an explicit DB lifecycle contract: block new DB work, pause/cancel scheduled DB writers, wait for active writes, then single-flight close the one handle. If the crash stack confirms Expo #38168, add `finalizeUnusedStatementsBeforeClosing: false` as part of that lifecycle fix.

PR #477, opened 2026-05-22
fix(app): consolidate historical transfer leftovers

vitalyiegorov:
## What changed

- Added historical consolidation candidates for past transfer leftovers that were missed by the newer watch-case logic.
- Detects FOP EUR -> FOP UAH bridge leftovers that should fold into an existing card transfer.
- Detects closest same-currency income duplicates that should attach to an existing transfer.
- Stabilized PrivatBank XLS import identity by hashing raw statement fields instead of parsed `Date#toISOString()`.
- Added `scripts/repair-privatbank-duplicate-imports.sql` for a one-time soft-delete repair of existing duplicate PrivatBank imports in a DB dump.
- Added bank-sync regression coverage for historical consolidation and PrivatBank identity stability.

## Why

The recent consolidation paths handled new/watch cases, but older imported rows often lacked the same IBAN hints or had bank-side time offsets. Separately, PrivatBank XLS rows could be reimported as new transactions when the same raw statement date parsed to a different JS timestamp, because the generated external ID depended on `toISOString()`.

## Validation

- `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/privatbank/external-id.test.ts`
- `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/consolidation/monobank-privatbank-transfer.test.ts src/scenarios/privatbank/external-id.test.ts`
- `yarn workspace @budgie-at/bank-sync-tests test`
- `yarn workspace @budgie/bank-sync build`
- `yarn format`
- `yarn ts`
- `yarn lint` (passes with the existing `use-suggestion-base.hook.ts` exhaustive-deps warning)
- `yarn deadcode`
- `yarn cpd`
- `git diff --check`
- Repair SQL tested on a copied simulator DB: selected 132 duplicate PrivatBank imports and left 0 visible one-hour duplicate pairs in the copy.


packages/app/src/sync/service/transfer-consolidation.service.ts
liaugust, line 1: consider splitting the service to smaller ones

liaugust (changes requested):

vitalyiegorov:
Addressed the service-size review with a smaller split and then trimmed the over-split pieces:

- kept consolidation orchestration, candidate discovery, auto-candidate processing, and executor responsibilities separated
- removed the single-consumer source-id and writer services
- kept the executor logs/write path together so historical consolidation failures still have enough trace data
- validation rerun locally:
  - `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/consolidation/historical-transfer-leftovers.test.ts`
  - `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/consolidation`
  - `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd`

Lint still emits the existing `use-suggestion-base.hook.ts` exhaustive-deps warning, but exits successfully.

vitalyiegorov:
Merged the non-duplicative part of #486 into this PR in commit `3eb5708fa` (`docs: document simulator dev testing`). I intentionally did not keep the extra legacy bridge SQL/test from #486 because this branch already has `EXISTING_TRANSFER_BRIDGE_CANDIDATES_SQL` plus `historical-transfer-leftovers.test.ts` covering the same simulator shape; keeping both made the same case consolidate twice.

Fresh local validation after integration:
- `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/consolidation` -> 19 files, 42 tests passed
- `yarn ts` -> 7 packages successful
- `git diff --check` -> clean

packages/app/src/@generic/service/foreground-workload.service.ts
```diff
@@ -0,0 +1,50 @@
+import { emptyFn } from '@rnw-community/shared';
+
+class ForegroundWorkloadService {
+    private activeCount = 0;
+    private readonly listeners = new Set<() => void>();
+
+    isActive(): boolean {
+        return this.activeCount > 0;
```
liaugust: isPositiveNumber

PR #486, opened 2026-05-23
fix(contracts): consolidate legacy bridge transfers

vitalyiegorov:
## What changed

- Adds a legacy IBAN bridge transfer candidate path for old imported pairs where source/bridge entries do not have `to_iban` values.
- Unions that path with the existing IBAN bridge transfer candidate query.
- Allows nullable bridge candidate IBAN fields in the shared candidate interface.
- Adds a regression fixture for the real legacy bridge-to-target transfer shape.
- Documents serve-sim dev testing rules in `AGENTS.md`.

## Why

Past Monobank FOP EUR -> FOP UAH -> card chains were not picked up by the newer IBAN-based bridge consolidation because historical entries lacked the entry IBAN metadata. The new candidate path matches those rows through the already-created bridge-to-target direct transfer, same-bank account IBAN prefix, exchange-rate tolerance, and the existing fast time window.

## Validation

- `yarn workspace @budgie-at/bank-sync-tests test src/scenarios/consolidation` -> 18 files, 41 tests passed
- `yarn ts` -> 7 packages successful
- `git diff --check` -> clean
- Dev build via serve-sim on iPhone 17 Pro: consolidation returned `0` high-confidence pairs and `50` review-only pairs after the real-data consolidation pass; no spinner hang
- Live simulator DB: `IBAN_BRIDGE_TRANSFER` total `79`, new bridge consolidations `id >= 79000` count `78`, January screenshot rows parented `9/9`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -478,6 +478,17 @@ Free-form `context: string`. Convention: hook/file/component name. Instantiate o
 4. **Before PR:** Run all validation commands
 5. **Do not commit new Markdown notes from agent work unless explicitly requested.** If a local instruction, scratch note, report, or generated Markdown file is needed only for the working session, keep it untracked and add the local pattern to `.gitignore` instead of committing it.
 
+## Simulator Dev Testing
+
+1. For dev-client feature checks and debugging, use the local `serve-sim` skill and the Codex in-app browser.
+2. Read `.agents/skills/serve-sim/SKILL.md` before using serve-sim. Follow its referenced workflow files when interacting with the simulator.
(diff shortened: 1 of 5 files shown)


vitalyiegorov:
Closing this draft as merged into #477. During integration, #477 already contained the historical bridge consolidation implementation and regression coverage, so the duplicate SQL/test changes from this branch were not kept. The non-overlapping `AGENTS.md` serve-sim dev-testing instructions were added to #477 in commit `3eb5708fa`.

PR #536, opened 2026-06-02
fix(app): refresh Monobank background sync registration

### LINKED ISSUE #532: Monobank auto sync does not run in background
## Problem
Monobank automatic sync does not run when the app is woken in the background.

## Expected
Registered background tasks should load their TaskManager definitions during headless launches so Monobank sync can execute without the rendered app layout.

## Fix direction
Move background task side-effect imports to the app entrypoint before expo-router/entry.

liaugust:
Follow-up to #534 after dev-client testing showed background sync still did not run.

Summary:
- Re-register Monobank background task when already present so stale dev-client registrations pick up the 15-minute interval.
- Register background tasks in sequence with Monobank last because Expo background tasks share one worker interval.
- Store and read the PIN with AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY so locked-device background wakes can open the encrypted database after first unlock.

Testing:
- yarn format
- yarn ts
- yarn lint (passes with existing use-suggestion-base exhaustive-deps warning)
- yarn deadcode
- yarn cpd
- yarn workspace @budgie-at/bank-sync-tests test src/scenarios/monobank

Fixes #532

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable PIN storage/retrieval with consistent secure-store options, improving background access and task registration.
  * Background sync/registration now consistently (re)registers tasks to avoid missed jobs.

* **New Features**
  * Added a workload runner to serialize and supervise background/foreground sync jobs.

* **Refactor**
  * Startup initialization now defers and sequences service syncs and task registration; background registration is gated by PIN accessibility.

* **Chores**
  * Several exported timing constants replaced with internal interval values for simpler scheduling.

### DIFF
--- a/packages/app/src/@generic/hook/use-app-initialization.hook.ts
+++ b/packages/app/src/@generic/hook/use-app-initialization.hook.ts
@@ -8,6 +8,7 @@ import { accountBalanceIncrementalService } from '../../account/service/account-
 import { authService } from '../../auth/service/auth.service';
 import { exchangeRatesSyncService } from '../../exchange-rate/service/exchange-rates-sync.service';
 import { monobankSyncService } from '../../sync/service/monobank-sync.service';
+import { syncWorkloadService } from '../../sync/service/sync-workload.service';
 import { transferConsolidationService } from '../../sync/service/transfer-consolidation.service';
 
 const logger = getLogger('useAppInitialization');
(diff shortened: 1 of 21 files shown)


packages/app/src/sync/task/transfer-consolidation.task.ts
```diff
@@ -5,14 +5,17 @@ import * as TaskManager from 'expo-task-manager';
 import { getErrorMessage } from '@rnw-community/shared';
 
 import { TRANSFER_CONSOLIDATION_TASK } from '../constant/transfer-consolidation-task.constant';
+import { syncWorkloadService } from '../service/sync-workload.service';
 import { transferConsolidationService } from '../service/transfer-consolidation.service';
 
 const logger = getLogger('transferConsolidationTask');
```
vitalyiegorov: I would have removed this loggers as we have logging on services itself

packages/app/src/@generic/hook/use-app-initialization.hook.ts
```diff
@@ -19,25 +21,30 @@ const runInitializationTask = (task: Promise<unknown>): void => {
     void task.catch(handleInitializationError);
 };
 
+const registerBackgroundTasks = async (): Promise<void> => {
```
vitalyiegorov: Inline

packages/app/src/@generic/hook/use-app-initialization.hook.ts
```diff
@@ -5,8 +5,10 @@ import { useEffect } from 'react';
 import { getErrorMessage } from '@rnw-community/shared';
 
 import { accountBalanceIncrementalService } from '../../account/service/account-balance-incremental.service';
+import { authService } from '../../auth/service/auth.service';
 import { exchangeRatesSyncService } from '../../exchange-rate/service/exchange-rates-sync.service';
 import { monobankSyncService } from '../../sync/service/monobank-sync.service';
+import { syncWorkloadService } from '../../sync/service/sync-workload.service';
 import { transferConsolidationService } from '../../sync/service/transfer-consolidation.service';
 
 const logger = getLogger('useAppInitialization');
```
vitalyiegorov: Sync service now has logs, do we need them here?

vitalyiegorov (changes requested):

PR #546, opened 2026-06-06
fix(app): reduce Home tab SQLite fanout

vitalyiegorov:
## Summary

- consolidate Home balance, section total, asset chip, and card balance reads into one Home balance summary live query
- pass balances through Home props so account cards and headers no longer mount their own SQLite subscriptions
- move Drizzle Studio out of Home and gate it behind `EXPO_PUBLIC_DRIZZLE_STUDIO_ENABLE` in dev
- isolate startup and foreground sync failures so one sync step does not block the rest
- leave React `Activity` behavior unchanged

## Validation

- `rtk yarn format`
- `rtk yarn ts`
- `rtk yarn lint` (passes with the existing `use-suggestion-base` exhaustive-deps warning)
- `rtk yarn deadcode`
- `rtk yarn cpd`

## Summary by CodeRabbit

* **New Features**
  * Drizzle Studio integration for development (toggleable via environment); env template updated.
* **Bug Fixes**
  * Foreground data sync now suppresses individual step errors for more resilient startup/background sync.
* **Refactor**
  * Balance and net-worth data flow centralized into a single summary hook and passed down to UI.
  * Many UI components now receive precomputed balances and counts instead of fetching them individually.


packages/app/src/account/component/account-grid-row/account-grid-row.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface Props {
+    readonly row: AccountRowInterface;
+    readonly accountType: AccountTypeEnum;
+    readonly balancesByAccountId: ReadonlyMap<number, HomeAccountBalanceInterface>;
+    readonly className?: string;
+}
+
+export const AccountGridRow = ({ row, accountType, balancesByAccountId, className }: Props) => {
+    const leftBalance = balancesByAccountId.get(row.left.id)?.balance ?? 0;
+    const rightAccount = row.right;
```
liaugust: Use just row.right instead of rightAccount

packages/app/src/account/query/use-accounts-with-bank-sync.query.ts
```diff
@@ -1,14 +1,12 @@
...
 
-import { isDefined } from '@rnw-community/shared';
-
 import { accountRepository } from '../../@generic/drizzle/db/db';
 
 export const useAccountsWithBankSyncQuery = () => {
     const { data, updatedAt, error } = useLiveQuery(accountRepository.findAllWithBankSync(), []);
 
-    const accounts: AccountWithBankSyncEntityInterface[] = isDefined(data) ? data : [];
+    const accounts: AccountWithBankSyncEntityInterface[] = data;
```
liaugust: I think we don't need this anymore

liaugust (changes requested):

liaugust (approved):

PR #561, opened 2026-06-07
feat: add Binance account sync with bank↔P2P consolidation

liaugust:
## Summary

Adds read-only **Binance account sync** and **bank↔Binance-P2P consolidation**, plus crypto amount/fee display formatting.

### 1. Binance sync (`bank-sync` + `app`)
- New `AccountTypeEnum.CRYPTO_SYNC`; HMAC-SHA256 signed client (`@noble/hashes`); creds packed in `bank_syncs.token`.
- **Bank-authoritative balances** (the Binance-reported snapshot, with Flexible + Locked Simple Earn folded; sold-off assets anchored to 0 — never ledger-derived).
- Transaction history: **spot trades & Convert** as cross-asset transfers (net-worth preserving), **P2P/C2C** income/expense, **Simple Earn rewards aggregated monthly per asset**, deposits/withdrawals.
- Backfill: windowed/dormancy paging (deposits 90-day chunks, earn ~30-day), weight-header throttling, incremental per-type commit (survives the iOS background kill), `exchangeInfo` symbol validation, 429/-1003 retry + pacing.
- Valuation reuses the existing CoinGecko crypto pipeline; Binance asset→instrument via a full-coverage alias map; unresolved assets park.

### 2. Bank ↔ P2P consolidation (`contracts` + `consolidation`)
- New `AUTO_P2P_FIAT_TRANSFER` bucket auto-merges a bank fiat **expense** with the matching Binance **P2P** crypto income (and the reverse for cash-outs) into one cross-currency `TRANSFER`.
- **Crypto↔fiat rate support**: the transfer-pair matcher now **triangulates exchange rates through the base instrument**, so UAH↔USDT (any crypto↔fiat) resolves even with no direct rate.
- Match gate: `binance:c2c:` leg ↔ non-Binance bank leg, implied rate within 10% of the triangulated market rate, within a 1h window. New `P2P_FIAT_TRANSFER` consolidation type (text enum, no migration).

### 3. Crypto display (`app`)
- Transaction-card amounts and the fee badge render crypto with up to **6 decimals** in `{amount} {symbol}` form (e.g. `0.278412 USDT`, `Fee 0.000065 BNB`); fiat keeps `{symbol}{amount}`.

## Verification
- `yarn format / ts / lint / deadcode / cpd` — all green.
- **bank-sync-tests 156**, **consolidation-tests 42** passing (incl. a new BUY/SELL/off-rate/off-window P2P scenario).
- Consolidation logic replicated against a real device DB: **50 P2P incomes → 50 correct consolidations, 0 false matches**.

## Notes / follow-ups
- Activation in-app: rebuild `contracts`+`bank-sync`+`consolidation`, restart Metro `-c`; consolidation works on already-synced data (no resync needed).
- Known deferred items tracked in the linked issue (per-symbol `fromId` cursor; "Clear all data" fix).

## Summary by CodeRabbit

* **New Features**
  * Binance account sync: connect with read‑only API key to sync balances, trades, transfers, deposits/withdrawals and earn rewards; two‑step setup UI (credentials → account selection) and parked‑assets notice.
  * New crypto-sync account type and Binance-backed create‑account option.

* **Improvements**
  * Crypto‑sync accounts reflected in net‑worth and balance displays; account icons/colors updated.
  * Improved fee/amount formatting for crypto transactions.

* **Localization**
  * Added Binance UI copy for DE/EN/ES/FR/UK translations.


liaugust:
Follow-ups / known limitations tracked in #562.

packages/app/src/account/service/account-balance-incremental.service.ts
```diff
@@ -18,7 +19,7 @@ class AccountBalanceIncrementalService {
         (error, truncate, tx) => `throw truncate=${String(truncate)} tx=${String(isDefined(tx))} error=${getErrorMessage(error)}`
     )
     async updateAllBalances(truncate: boolean, tx?: DB): Promise<void> {
-        const accounts = await accountRepository.getAllActiveAccounts(tx);
+        const accounts = (await accountRepository.getAllActiveAccounts(tx)).filter(account => account.type !== AccountTypeEnum.CRYPTO_SYNC);
```
liaugust: Include the filter logic into the db request inside the repository

packages/app/src/account/service/account-balance-incremental.service.ts
```diff
@@ -34,7 +35,9 @@ class AccountBalanceIncrementalService {
             return;
         }
 
-        const accounts = await accountRepository.findByIds(uniqueAccountIds, tx);
+        const accounts = (await accountRepository.findByIds(uniqueAccountIds, tx)).filter(
+            account => account.type !== AccountTypeEnum.CRYPTO_SYNC
+        );
```
liaugust: Include the filter into the db request inside the repository

packages/app/src/app/(main)/create-account/index.tsx
```diff
@@ -70,6 +70,12 @@ export default function Index() {
                     route={`/create-account/${ExternalSourceEnum.ERSTE}`}
                     bankProvider={ExternalSourceEnum.ERSTE}
                 />
+                <CreateBankSyncCard
```
liaugust: I think we should rename the component to be generic

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,698 @@
...
+        token => `enter keyLen=${token.length}`,
+        (result, token) =>
+            `done keyLen=${token.length} externalIds=${result.map(preview => preview.externalId).join(',')} parkedCount=${result.filter(preview => preview.isParked).length}`,
+        (error, token) => `throw keyLen=${token.length} error=${getErrorMessage(error)}`
+    )
+    async fetchAccountsPreview(token: string): Promise<BankAccountPreviewInterface[]> {
+        const bankAccounts = await this.fetchBankAccounts(token);
+
+        return this.mapAccountsToPreview(bankAccounts);
+    }
```
liaugust: It is not bank, it is crypto exchange

packages/app/src/sync/service/binance-sync.service.ts
liaugust, line 1: private methods should not have @Log

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1315 @@
...
+        return [
+            ...depositTransactions,
+            ...withdrawalTransactions,
+            ...fiatDepositTransactions,
+            ...fiatWithdrawalTransactions,
+            ...c2cTransactions,
+            ...earnRewardTransactions
+        ]
+            .sort((left, right) => right.time - left.time)
+            .slice(0, MAX_TRANSACTIONS_PER_WINDOW);
```
vitalyiegorov: **Vital: silent permanent data loss for assets with >500 lifetime source items.**

`.slice(0, MAX_TRANSACTIONS_PER_WINDOW)` keeps the newest 500 merged items *per asset* — but here the "window" is the entire history (the app anchors the source window at the earliest Binance transaction). For an active P2P trader (e.g. USDT with >500 c2c orders), the oldest items are dropped after the sort, and since every subsequent run re-applies the same cap to the same full window, they are never imported. No log, no error — the ledger is just quietly incomplete.

`MAX_TRANSACTIONS_PER_WINDOW` is the deposit-endpoint page limit; reusing it as a cap on the merged in-memory list serves no API constraint. Suggest dropping the `.slice(...)` entirely (the upstream fetches are already windowed/paged), or at minimum logging loudly when the cap truncates.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1315 @@
...
+            error.message.includes(String(UNAUTHORIZED_STATUS)) ||
+            error.message.includes(String(FORBIDDEN_STATUS));
+        if (isPermissionDenied) {
+            syncLogger.log('binance:c2c:unavailable', { tradeType, message: error.message });
+
+            return this.success({ orders: [], hasMore: false });
+        }
+
+        return this.failure(SyncError.invalidResponse(this.provider));
+    }
```
vitalyiegorov, line 1206: **Vital: this swallows real failures and reports them as "no transactions".**

Two problems:

1. The fallback `return this.failure(SyncError.invalidResponse(this.provider))` discards the original error code. A persistent 429 (`RATE_LIMITED`) or network error during C2C paging becomes `INVALID_RESPONSE`, and the app side (`resolveSourceTransactions` in `binance-sync.service.ts`) maps `INVALID_RESPONSE` → `[]`. Net effect: a rate-limited/failed C2C fetch completes the sync "successfully" with zero C2C transactions, no error recorded, no backoff. The original result should be propagated: `return result;` / `return this.failure(error)`.

2. `error.message.includes(String(UNAUTHORIZED_STATUS))` is substring matching on `'401'`/`'403'` — it will match any message that happens to contain those digit sequences (amounts, timestamps, order ids). The root cause is that `BaseSyncProviderClient.handleError` maps 403 to `UNKNOWN` with a `HTTP 403: ...` message. Cleaner fix: map 403 → `UNAUTHORIZED` (or a dedicated `FORBIDDEN` code) in the base client and check `error.code` here, per the canonical-guard spirit of CLAUDE.md — no string sniffing.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1315 @@
...
+                baseAssets.add(balance.asset);
+            }
+        }
+
+        for (const transfer of convertTransfers) {
+            this.addTransferAssets(baseAssets, transfer);
+        }
+
+        return baseAssets;
+    }
```
vitalyiegorov, line 917: **Inconsistency: trades of fully sold-off assets are never discovered.**

Symbol candidates come only from (a) current positive balances and (b) Convert legs. An asset that was bought and fully sold via spot trades (no convert, zero balance today) never produces a candidate symbol, so its `myTrades` history is never fetched — while its deposits/withdrawals *are* imported (those discover assets from history). Result: the ledger contains the deposit but not the disposal, even though the PR description says trades are net-worth-preserving transfers. Balances stay correct only because of the anchor-to-0 behavior, but the history is silently asymmetric.

If this is an accepted limitation (Binance offers no "all my traded symbols" endpoint, and probing every symbol is weight-prohibitive), it's worth a note in the PR description / linked issue. Alternatively, assets seen in deposit/withdrawal/c2c history could be added to `baseAssets` here — they're already fetched.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1315 @@
...
+        const free = binanceMapper.parseBinanceAmount(balance.free);
+        const locked = binanceMapper.parseBinanceAmount(balance.locked);
+        if (!isDefined(free) || !isDefined(locked)) {
+            return null;
+        }
+
+        const total = free + locked;
+
+        return total * MICRO_UNITS_PRECISION > Number.MAX_SAFE_INTEGER ? null : total;
+    }
```
vitalyiegorov: **Duplication:** `computeTotalBalance` re-implements `computeRawTotalBalance` line-for-line and only adds the overflow check — it should call it:

```ts
private computeTotalBalance(balance: BinanceAssetBalanceApiInterface): number | null {
    const total = this.computeRawTotalBalance(balance);

    return isDefined(total) && total * MICRO_UNITS_PRECISION <= Number.MAX_SAFE_INTEGER ? total : null;
}
```

Same theme elsewhere in this file: `resolveSourceWindow` exists, but `getTransactions` (lines 156–157) and `getTransfers` (lines 260–261) re-inline the same seconds→ms math with subtly different "now" derivations (`Math.floor(Date.now() / 1000) * 1000` vs `Date.now()`). One helper, one definition of "now".

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,505 @@
...
+        this.runSignedClient = null;
+        this.runClientToken = null;
+    }
+
+    private async resolveSourceWindowStart(): Promise<number> {
+        const earliestTransactionTime = await transactionService.getEarliestTransactionTimeByExternalSource(this.provider);
+        const from = subYears(new Date(), BINANCE_TRANSFER_LOOKBACK_YEARS);
+
+        return getUnixTime(earliestTransactionTime ?? from);
+    }
```
vitalyiegorov: **Vital (perf/API budget): every forward poll re-walks the entire history.**

The source window start is anchored at the *earliest* Binance transaction (or 5-year lookback), not at the last-synced point. With the 15-minute background task plus the 2-minute foreground staleness threshold, every run re-fetches:

- C2C: ~61 thirty-day windows × 2 trade types (dormancy is `UNBOUNDED`)
- Earn rewards: ~61 windows; Convert: ~61 windows; deposits/withdrawals: ~20 ninety-day windows × 2
- `myTrades` from `fromId=0` per candidate symbol (full history, every run — I know the cursor is a tracked deferral)

That's hundreds of signed requests per poll, forever, growing with account age — most returning data that's then discarded by the existing-externalId filter. The throttle turns this into multi-minute runs (60 s cool-downs) inside an iOS background task budget of ~30 s.

Was anchoring at the earliest transaction intentional (e.g. to self-heal the C2C/trade gaps)? If not, forward runs should start at `forwardSyncedAt`/last-synced minus a small overlap, with the full walk reserved for the initial backfill. If it *is* intentional, that trade-off deserves a note, because it also masks the data-loss issue flagged in `buildTransactions`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,505 @@
...
+
+    private getRunSignedClient(token: string): BinanceSignedClient {
+        if (!isDefined(this.runSignedClient) || this.runClientToken !== token) {
+            this.runSignedClient = new BinanceSignedClient(token);
+            this.runClientToken = token;
+        }
+
+        return this.runSignedClient;
+    }
```
vitalyiegorov: **Two `BinanceSignedClient` instances per run + uncached account fetches.**

`getRunSyncService` builds a `BinanceSyncService`, which constructs its *own* `BinanceSignedClient` internally, while `getRunSignedClient` builds a second one. The two don't share the per-run caches, server-time offset, or `validSymbols` — and `BaseSyncService` adds nothing here beyond `syncAccounts()` (the forward/backward paging it provides is unused by the snapshot model). One run-scoped `BinanceSignedClient` would do; `fetchBankAccounts` can call `client.getAccounts()` directly.

Related: `fetchBankAccounts` is a full balance pipeline (`getUserAsset` + flexible-earn pagination + locked-earn pagination + funding wallet ≈ 4+ signed calls) and is invoked up to ~6× per run — `anchorAllBalances`, once per `createSyncedSources` source type, and `createSyncedTransfers` — plus `deriveTradeSymbols` does its own `getUserAsset`. The client meticulously caches every history endpoint per run, but the single most-repeated call has no cache. Caching the accounts result per run (it's the same snapshot by definition) would cut a large fraction of the per-run request weight.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,505 @@
+/* eslint-disable no-await-in-loop, max-lines -- Sync orchestration requires sequential awaits and the account-creation loop reads the running account count; the file owns one cohesive sync service */
```
vitalyiegorov: **CLAUDE.md: file-wide disable + justification doesn't match the file.**

1. Repo convention is per-line `eslint-disable-next-line` with a justification (that's how `binance-signed.client.ts` handles its `no-await-in-loop` cases); a file-wide `no-await-in-loop` disable also silences future loops that *should* be parallelized.
2. The justification text — "the account-creation loop reads the running account count" — doesn't describe anything in this file (`createdCount` is a plain counter; no read of a running count gates the loop).
3. Per rules 4/38, `max-lines` disables require explicit human approval; the client and mapper files carry the `-- approved by human` marker, this one doesn't. Inconsistent within the same PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/transaction/service/transaction.service.ts
```diff
@@ -264,6 +247,116 @@ class TransactionService {
...
 
+    private async persistPrimaryTransfer({
+        transaction,
+        input,
+        fromEntry,
+        toEntry,
+        fromAmountInMicroUnits,
+        toAmountInMicroUnits,
+        tx
+    }: PersistPrimaryTransferInputInterface): Promise<void> {
```
vitalyiegorov: **CLAUDE.md rules 33/42: five new single-consumer param-bag interfaces to dodge `max-params`.**

`PersistPrimaryTransferInputInterface`, `PersistTransferInputInterface`, `ValueTransferLegsInputInterface`, `TransferLegValuationsInterface`, `ValuedTransferEntryInterface` are all one-off shapes consumed by exactly one private method each, and they exist to keep these signatures under the params limit — including smuggling `tx: DB` inside the bag. Rule 33 says don't reshape arguments into objects/interfaces for lint; rule 42 says don't resolve lint pressure with one-off interfaces; the prescribed escape hatch is a narrow `@typescript-eslint/max-params` disable with justification on the private method.

The extraction of `persistPrimaryTransfer`/`valueTransferLegs` itself is good (it's what lets `createSyncedTransfer` reuse the internal-transfer path) — it's only the parameter shape that fights the house rules. Positional args + a targeted disable would drop 5 interface files.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sync/src/binance/constant/binance-asset-instrument-code.constant.ts
```diff
@@ -0,0 +1,104 @@
+import { BinanceAssetEnum } from '../enum/binance-asset.enum';
+
+export const BINANCE_ASSET_INSTRUMENT_CODE: Record<BinanceAssetEnum, string> = {
```
vitalyiegorov: **Overengineering: ~100-member enum + ~100-entry record where ~95% of entries are `X → 'X'` identity.**

The only information this pair encodes is a handful of real aliases (`MATIC → POL`, …). Everything else is ceremony that must be hand-extended for every new Binance listing — and until someone ships that app update, a user's newly listed asset parks even when the instrument already exists in the DB (the actual gate is `instruments.find(code === instrumentCode)` on the app side).

Suggest inverting it: a small `BINANCE_ASSET_ALIAS: Record<string, string>` of the genuine exceptions, with `resolveBinanceInstrumentCode(asset)` falling back to `asset` itself. The instrument repository remains the real authority on what's supported; unknown assets still park exactly as today. That deletes ~200 lines and removes the standing maintenance burden, with strictly better coverage.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/util/map-binance-transfer-to-create-input.util.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import { isDefined, isPositiveNumber } from '@rnw-community/shared';
+
+import type { AccountEntityInterface, TransactionCreateInputInterface, TransactionEntryCreateInputInterface } from '@budgie/contracts';
+
+const FEE_ENTRY_EXTERNAL_ID_SUFFIX = ':fee';
+const MILLISECONDS_PER_SECOND = 1000;
+
+export const mapBinanceTransferToCreateInput = async (
+    transfer: BinanceTransferInterface,
```
vitalyiegorov: **CLAUDE.md rule 52: single-consumer util.**

Both this file and `resolve-binance-instrument-code.util.ts` are imported by exactly one consumer (`AppBinanceSyncService`). Per rules 38/42/52, single-consumer helpers belong inline as private methods of the consuming class (`/utils` is reserved for genuinely shared pure helpers — `map-bank-transaction-to-create-input.util.ts` qualifies because Monobank shares it; these two don't).

While moving it: when `transfer.feeAmount > 0` but the fee asset's account can't be resolved (parked instrument), `hasFee` goes false and the fee leg is *silently* dropped from an otherwise-created transfer — unlike the from/to legs, which skip the whole transfer with a `skip-parked-leg` log. Either log the dropped fee or fall back to `fromAccount` like the `feeAccount?.id ?? fromAccount.id` line already implies was considered.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/contracts/src/account/enum/account-association.enum.ts
```diff
@@ -4,5 +4,5 @@ export enum AccountAssociationEnum {
     INSTRUMENT = 'instrument',
     BALANCES = 'balances',
     PARENT = 'parent',
-    BANK_SYNC = 'bankSync'
+    BANK_SYNC = 'sync'
```
vitalyiegorov: **Inconsistency:** the rename updated the value but not the key — `BANK_SYNC = 'sync'` now lies about itself, and the rest of this PR consistently renamed the BankSync vocabulary to Sync (`SyncEntityTable`, `SyncRepository`, `AccountAssociationEnum` consumers…). Should be `SYNC = 'sync'`; rule 28's "preserve the serialized value" exception covers the value, not a stale key.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/contracts/src/account/repository/account.repository.ts
```diff
@@ -55,6 +58,13 @@ export class AccountRepository {
         return await (tx ?? this.db).select().from(AccountEntityTable).where(isNull(AccountEntityTable.deletedAt));
     }
 
+    async getAllActiveAccountsExceptBankAuthoritative(tx?: DB): Promise<AccountEntityInterface[]> {
+        return await (tx ?? this.db)
+            .select()
+            .from(AccountEntityTable)
+            .where(and(isNull(AccountEntityTable.deletedAt), ne(AccountEntityTable.type, AccountTypeEnum.CRYPTO_SYNC)));
+    }
+
```
vitalyiegorov, line 67: **"Bank-authoritative" is now hardcoded as `type = CRYPTO_SYNC` in four places.**

This method, `findByIdsExceptBankAuthoritative` below, `AccountBalanceRepository.truncateExceptBankAuthoritative`, and the `CASE WHEN accounts.type = CRYPTO_SYNC THEN 0` ledger-sum guard all independently encode the same business rule ("this account type's balance comes from the provider snapshot, not the ledger"). The next bank-authoritative provider (a future exchange, or a bank that reports authoritative balances) means hunting all four down.

Suggest a single `BANK_AUTHORITATIVE_ACCOUNT_TYPES` constant (or an `isBankAuthoritative` helper) in contracts that all four sites consume — same pattern as `AccountBalanceRepository.CRYPTO_ACCOUNT_TYPES` introduced in this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/transaction/components/transaction-amount/transaction-amount.tsx
```diff
@@ -63,19 +64,27 @@ const getDisplayState = (transaction: TransactionWithRelationsEntityInterface) =
...
+    const formatEntryAmount = (entry: AggregatedEntry): string => {
+        const amount = convertFromMicroUnits(entry.amount);
+        const instrumentSymbol = entry.account.instrument.symbol;
+
+        if (entry.account.instrument.type === InstrumentTypeEnum.CRYPTO) {
+            return `${formatCryptoDigits(amount)} ${instrumentSymbol}`;
+        }
+
+        return formatDigits(amount, instrumentSymbol);
+    };
```
vitalyiegorov: **Duplication:** the crypto-vs-fiat formatting branch (`isCrypto ? \`${formatCryptoDigits(amount)} ${symbol}\` : formatDigits(amount, symbol)`, plus the paired `useFormatDigits(0, MICRO_UNIT_DECIMAL_PLACES)` setup) now lives both here and in `TransactionFeePill`. With two consumers this meets the bar for a shared helper (rule 14/38) — e.g. a `useFormatInstrumentAmount()` hook that takes `(amount, instrument)` and owns the crypto/fiat decision, so the next surface that renders crypto amounts (account cards, statistics) doesn't add a third copy with slightly different rules.

---
_Generated by [Claude Code](https://claude.ai/code)_

vitalyiegorov (commented):
Deep review of the Binance sync + P2P consolidation work (requested by @vitalyiegorov). Overall the architecture is solid — the sync-service hierarchy refactor, Zod-validated API boundaries, per-run caching, weight throttling, and the SQL rate triangulation for the P2P matcher are all well built, and the test coverage in `tests/sync-tests` / `tests/consolidation-tests` is genuinely good. Inline comments cover the substantive findings; summary below.

## Vital flaws
1. **Silent permanent data loss**: `buildTransactions` caps merged history at 500 items per asset while the fetch window is the entire history — oldest P2P/earn/deposit items beyond 500 are never imported, silently (`binance-signed.client.ts`).
2. **Swallowed failures**: `handleC2cFailure` rewrites `RATE_LIMITED`/network errors as `INVALID_RESPONSE`, which the app layer maps to "no transactions" — a failing C2C fetch completes the sync successfully with missing data. Plus fragile `message.includes('401')` string matching.
3. **API budget**: every forward poll re-walks the full multi-year window set (source window anchored at the *earliest* transaction) and re-fetches `myTrades` from `fromId=0` — hundreds of signed requests per 15-min background run, growing with account age.
4. **Coverage gap**: trades of fully sold-off assets (no convert, zero balance) are never discovered, while their deposits are — asymmetric ledger.

## Overengineering / unnecessary complexity
5. ~100-member `BinanceAssetEnum` + identity-mapping record where an exceptions-only alias map with fallback to the asset code would be smaller and strictly more capable (new listings currently park until an app release).
6. Five single-consumer param-bag interfaces in the transaction module to dodge `max-params` — rules 33/42 prescribe positional args + a narrow disable.
7. Two `BinanceSignedClient` instances per run (one inside `BinanceSyncService`, one raw) with split caches/server-time, and the heavy accounts pipeline fetched ~6× per run with no per-run cache.

## CLAUDE.md / inconsistencies
8. File-wide `eslint-disable no-await-in-loop, max-lines` in `binance-sync.service.ts` with a justification that doesn't match the file and no `approved by human` marker (the client/mapper files in this same PR carry it).
9. `AccountAssociationEnum.BANK_SYNC = 'sync'` — key/value mismatch after the rename.
10. "Bank-authoritative ⇔ CRYPTO_SYNC" hardcoded in four places across two repositories — wants one shared constant.
11. Crypto amount formatting duplicated between `TransactionAmount` and `TransactionFeePill`; single-consumer utils (`map-binance-transfer-to-create-input`, `resolve-binance-instrument-code`) violate rule 52.

## Minor (no inline comment)
- `BinanceWeightThrottle.paceRequestRate` delays once and proceeds regardless — the request-rate ceiling isn't actually enforced (should re-check after the delay).
- `getTransfers(accountId, …)` validates `accountId` then ignores it (returns all transfers) — signature implies per-account filtering that doesn't happen; rename or drop the param (rule 37).
- `BaseSyncProviderClient.handleError`'s `max-statements` disable is justified as "temporary diagnostic logs" — either they're permanent (reword) or they should go.
- `type SetupStep = 'credentials' | 'accounts'` is an inline type alias in a component file (rule 19, strictly).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/task/binance-sync.task.ts
```diff
@@ -0,0 +1,14 @@
...
+import { binanceSyncService } from '../service/binance-sync.service';
+import { syncWorkloadService } from '../service/sync-workload.service';
+
+TaskManager.defineTask(BINANCE_SYNC_TASK, async () => {
+    try {
+        return await syncWorkloadService.run('background-binance', () => binanceSyncService.sync());
+    } catch {
+        return BackgroundTask.BackgroundTaskResult.Failed;
+    }
+});
```
vitalyiegorov, line 18: **Vital (iOS): this run shape is structurally incompatible with the BGTaskScheduler budget.**

`expo-background-task` on iOS rides `BGAppRefreshTask` — ~30 seconds of wall time, not guaranteed, and iOS penalizes future scheduling for tasks that get expired. The Binance run inside this task can legally contain:

- a **60-second `setTimeout` cool-down** in `BinanceWeightThrottle.coolDown()` the moment any weight crosses 80% of ceiling — a *single* cool-down already exceeds the entire task budget;
- the unbounded window walk (see the round-1 comment) — minutes for an aged account;
- `microPause(1000)` per sync row × one row per asset.

The per-source-type commits mean a kill isn't data loss, which is good — but as designed, **the background task will essentially never complete a Binance run and will train iOS to stop scheduling it**, leaving foreground as the only effective sync path (see the comment in `root-layout-content.tsx`).

The structural fix is a **deadline, not a flag**: thread a `deadlineAt` (e.g. `Date.now() + 25_000` for background, `Infinity` for foreground) through `sync()` into the window walkers; when the deadline is near, return `completed: false` with the current checkpoint so `resolveProgressUpdate` resumes next run. The throttle should then *defer* (end the run cleanly) instead of sleeping 60 s inside a doomed task.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/app/root-layout-content.tsx
```diff
@@ -74,6 +75,7 @@ const isDrizzleStudioEnabled = __DEV__ && process.env[drizzleStudioEnvironmentVa
 const syncForegroundData = async (): Promise<void> => {
     await exchangeRatesSyncService.sync().catch(emptyFn);
     await monobankSyncService.sync().catch(emptyFn);
+    await binanceSyncService.sync().catch(emptyFn);
```
vitalyiegorov, line 88: **Vital (UX): the unbounded Binance engine now runs on the live app at every foreground + every cold start.**

`handleAppStateChange(true)` and the startup hook both enqueue this chain, and the 2-minute staleness gate means any app-open after a 2-minute pause triggers a *real* Binance run — in the same JS runtime the user is interacting with. Two compounding costs that Monobank never had:

1. **Run duration**: full-history window walk + potential 60 s throttle cool-downs (a user opening the app mid-cool-down has a sync "running" for minutes while `foregroundWorkloadService.isActive()` stays true).
2. **Per-asset sync rows**: Binance creates one sync row per asset, all going stale together. `processPendingSyncs` handles them strictly sequentially with `microPause(rateLimitMs)` = **1 s per row** — a 30-asset portfolio pays ~30 s of pure pacing per foreground run even though rows 2…N are no-ops (the `*ThisRun` flags already did all the work on row 1). The rate-limit pause exists to space *API calls*; a no-op row makes zero API calls.

Minimum viable fix: skip the pause (or short-circuit the remaining provider rows) once `executeSyncBatch` reports the run-level work is done. The deeper fix is one provider-level sync row for snapshot-model providers — the per-account row is a Monobank-ism that doesn't fit Binance's "one token, one snapshot, N assets" shape.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,505 @@
...
+            const isCreated = await this.createSyncedTransfer(transfer, resolveAccount);
+            createdCount += isCreated ? 1 : 0;
+        }
+
+        if (isPositiveNumber(createdCount)) {
+            await transactionService.updateAllBalances();
+        }
+
+        return createdCount;
+    }
```
vitalyiegorov, line 436: **Vital (perf + app CLAUDE.md): backfill write amplification, and the mandated batch+yield pattern is skipped.**

For an initial backfill of a few thousand trades, this loop does — per transfer — a full DB write transaction (`createSyncedTransfer` → `persistPrimaryTransfer` → `persistTransfer`), and `persistTransfer` ends with `updateBalancesByAccountIds(...)` **inside each per-transfer transaction**. So a 2 000-trade backfill executes 2 000 write transactions + 2 000 incremental balance recomputes + the final `updateAllBalances()` here — while running in the foreground JS runtime (see `root-layout-content.tsx`). The same pattern repeats in `commitSourceType`: `updateAllBalances()` after *each* of the four source types, plus once more at the end of `syncForegroundData` — up to **six full-ledger balance passes in one run**.

The app CLAUDE.md is explicit for exactly this case: *"Process in batches, commit each batch in its own short transaction, publish progress, then `await microPause()` before the next batch"* (the `RULE_BATCH_SIZE` pattern). Neither this loop nor `collectSourceInputs` yields or batches.

Suggested shape:
- chunk transfers (e.g. 50/batch), one DB transaction per chunk, `microPause()` between chunks;
- hoist the balance update out of the per-transfer path for the synced flow — collect touched account ids and run `updateBalancesByAccountIds` once per batch (or once per run);
- drop the per-source-type `updateAllBalances()` in favor of a single pass at the end of `executeSyncBatch` — the chain's trailing `updateAllBalances(false)` already exists as a safety net.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,505 @@
...
+        return anchoredCount;
+    }
+
+    protected override validateToken(token: string): void {
+        BinanceCredentialsSchema.parse(JSON.parse(token));
+    }
+
+    protected override async beforeSyncRun(): Promise<void> {
+        this.resetRunState();
+    }
```
vitalyiegorov: **Vital (memory): the full-history response caches are retained between runs.**

`resetRunState()` only runs in `beforeSyncRun()` — i.e. at the *start of the next* sync. Between runs, this singleton keeps `runSignedClient` (and `runSyncService`'s internal client) alive, and those clients hold every per-window cache populated during the run: all deposits, withdrawals, fiat orders, C2C orders, earn rewards, transfers, plus `validSymbols` (~thousands of strings from `exchangeInfo`). For an aged account that's the entire raw Binance history as parsed JSON, pinned in Hermes heap indefinitely on a memory-constrained device — and it's dead weight, because the next run resets it before use anyway.

`AbstractPollingSyncService.sync()` already has the right seam: its `finally` block. Add an `afterSyncRun()` hook (or call `resetRunState()` there for Binance) so the clients and their caches are released when the run ends, not when the next one starts.

---
_Generated by [Claude Code](https://claude.ai/code)_

vitalyiegorov (commented):
Round 2 — Expo/React Native runtime architecture pass, vital findings only. This round looked at how the sync engine actually behaves inside the two execution contexts it's wired into (BGTaskScheduler and the foreground JS runtime), not at code style.

**The common thread: the Binance engine is built as an unbounded batch job, but both of its hosts are time-and-resource-budgeted.**

1. **iOS background task can never finish it** — a single 60 s throttle cool-down exceeds the entire ~30 s `BGAppRefreshTask` budget; iOS will progressively stop scheduling the task. Needs a deadline threaded into the run with clean `completed: false` checkpointing (the per-source-type commits already make resume safe).
2. **Foreground runs it on the live app** — every app-open after 2 minutes triggers a real run in the interactive runtime, and the per-asset sync-row model adds ~1 s of pure pacing per asset even for no-op rows.
3. **Backfill write amplification** — per-transfer DB transactions each containing an incremental balance update, plus up to six full-ledger balance passes per run; the app's own mandated batch + `microPause` yield pattern isn't applied to the synced-transfer/source loops.
4. **Heap retention** — the run-scoped clients (holding the entire parsed Binance history in per-window caches) stay referenced between runs because cleanup happens in `beforeSyncRun` of the *next* run instead of a `finally`/`afterSyncRun`.

Checked and deliberately *not* flagging: SYNCING rows orphaned by an iOS kill are recovered (pending queries don't filter on status); credentials-at-rest are fine (SQLCipher DB, backups export the encrypted file); `defineTask` registration and result propagation are correct; the setup-flow hook has no stale-closure issue; the error-threshold self-disable behavior is pre-existing shared semantics, not new to this PR.

All four items compound with round-1 finding #3 (full-history re-walk per poll): fixing the window anchoring shrinks the constant factor everywhere, but the deadline/batching/cleanup shape is worth getting right regardless, since it's what every future snapshot-model provider will inherit.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/transaction/service/transaction.service.ts
liaugust, line 1: Add @Log to all public methods

packages/sync/src/monobank/mapper/monobank-account-type.mapper.ts
liaugust, line 1: why do we need this ?

tests/sync-tests/src/harness/binance/binance-server.ts
liaugust, line 1: we have to improve this

tests/sync-tests/src/harness/db/fetch-sync-by-id.ts
liaugust, line 1: rename

packages/sync/src/binance/mapper/binance.mapper.ts
```diff
@@ -0,0 +1,349 @@
...
+        return buckets;
+    }
+
+    private mapEarnMonthBucket(asset: string, accountId: string, bucket: BinanceEarnMonthBucket): SyncTransactionInterface | null {
+        if (bucket.totalMicroUnits <= 0) {
+            return null;
+        }
+
+        const amount = bucket.totalMicroUnits / BINANCE_MICRO_UNITS_PRECISION;
+        const externalId = `binance:earn:${asset}:${bucket.monthKey}`;
```
vitalyiegorov: **P0 — monthly Earn buckets freeze at their first-computed value; every reward received later in the same month is silently dropped.**

The bucket total sums only the rewards in the current fetch window, but this `externalId` is stable per `(asset, month)`. On the app side, existing ids are routed to `reconcileSourceAccounts` (`binance-sync.service.ts:302-314`), which only moves accounts and updates quote fields — **nothing ever updates `amount`** (`importedTransactionEntryUpdateService.update` is never called on the Binance path). Sync on Jul 3 writes `binance:earn:BTC:2026-07` = 3 days of rewards; every later sync computes a fresh partial bucket, finds the id exists, and skips it. The month permanently records only its first few days of staking income.

Fix: key buckets at day granularity (`binance:earn:${asset}:${YYYY-MM-DD}`) so a bucket is closed the moment it is written — this also removes the UTC-month-key skew in `buildMonthKey` — or route existing earn ids through an amount update with the earn fetch window floored at `startOfMonth(now)`.

packages/consolidation/src/query/repository/sql-factory/p2p-fiat-authoritative-candidate-sql.factory.ts
```diff
@@ -0,0 +1,148 @@
...
+            ON p2p_transaction.id = p2p_entry.transaction_id
+            AND p2p_transaction.deleted_at IS NULL
+            AND p2p_transaction.consolidation_parent_transaction_id IS NULL
+        INNER JOIN accounts p2p_account
+            ON p2p_account.id = p2p_entry.account_id
+            AND p2p_account.type = '${AccountTypeEnum.CRYPTO_SYNC}'
+            AND p2p_account.deleted_at IS NULL
+            AND p2p_account.is_active = 1
+        INNER JOIN instruments p2p_instrument ON p2p_instrument.id = p2p_account.instrument_id
+        WHERE p2p_entry.deleted_at IS NULL
```
vitalyiegorov, line 48: **P0 — `FEE` entries are not excluded from the P2P matcher; a fee-bearing bank expense (Monobank's default) breaks the bucket three ways.**

Neither `bank_entries`/`p2p_entries` here nor the expense/income CTEs in `transfer-pair-ranked-candidate-base-sql.factory.ts:64-67,107-109` filter `entry.type` — compare `refund-ranked-candidate-sql.factory.ts:60,92`, which gates `CREDIT`/`DEBIT`. A synced bank expense with commission has **two entries on the same account** (main + FEE, both `exchange_rate = 1`), and Monobank always populates `feeAmount`:

1. **Atomic BUY silently never consolidates** — grouping by `incomeTransactionId:expenseEntryAccountId` puts main+fee in one group; `collectExpenseCombinations` emits `[main, fee]` → `bankTransactionIds = [7, 7]`; that combination usually wins stage 1 (its sum is closest to the ideal), then stage 2 self-eliminates it (`owners = [c, c]` → tie → `null`). Zero candidates for that order, forever, no log.
2. **The bank fee vanishes from the ledger** — when the authoritative path matches the main entry alone, the canonical carries only the main amount; all moved source entries (incl. FEE) are ledger-excluded → balance short by `feeAmount`, fee expense gone from statistics. The ATM family solves exactly this via `createAtmCashWithdrawalFeeEntry`; P2P has no equivalent.
3. **Repair churn every run** — the repair count SQL counts *all* moved bank entries (= 2 for fee-bearing sources) against `!= 1`, so it unconsolidates, deletes and rebuilds the canonical on **every** consolidation pass, with a new canonical id each time.

Fix: add `AND type = 'CREDIT'`/`'DEBIT'` to the four entry CTEs, restrict the repair count to non-FEE entries, mint a canonical FEE entry the way ATM does, and harden `buildBuyCandidates` so a combination can never contain two rows from one `expenseTransactionId`. Related latent invariant: `buildExpenseCombinations`/`buildBuyCandidate` take income-side values from `rows[0]`/`combination[0]`, assuming one income entry per P2P transaction — true today only because C2C hardcodes `feeAmount: 0`.

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1460 @@
...
+        startTimeMs: number,
+        endTimeMs: number
+    ): Promise<SyncResultInterface<BinanceDepositApiInterface[]>> {
+        const cacheKey = `${wallet}:${DEPOSIT_HISTORY_ENDPOINT}:${startTimeMs}-${endTimeMs}`;
+        const cached = this.depositCache.get(cacheKey);
+        if (isDefined(cached)) {
+            return this.success(cached);
+        }
+
+        const result = await this.signedRequest(DEPOSIT_HISTORY_ENDPOINT, 'GET', {
```
vitalyiegorov, line 462: **P1 — deposit/withdrawal windows have no `offset` paging; >500 records in a 90-day window are silently dropped.**

`limit: MAX_TRANSACTIONS_PER_WINDOW` (500) with no `offset` anywhere in the package — both `/sapi/v1/capital/deposit/hisrec` and `/sapi/v1/capital/withdraw/history` support `offset` and `limit` up to 1000. `accumulateWindow` can't detect the truncation: a full window looks identical to a partial one, so the walk just moves on. Same failure shape as the fixed round-1 #1, on the API side.

Fix: page inside `fetchDepositWindow`/`fetchWithdrawalWindow` while `page.length === limit` (mirroring `fetchPagedOrders`), and raise `limit` to 1000.

packages/app/src/sync/service/binance-sync.service.ts
```diff
@@ -0,0 +1,611 @@
...
+        this.runSignedClient = null;
+        this.runClientToken = null;
+        this.runExchangeAccounts = null;
+    }
+
+    private isDeadlineDeferral(error: unknown): boolean {
+        return error instanceof SyncError && error.code === SyncErrorCodeEnum.RATE_LIMITED;
+    }
+
+    private async resolveSourceWindowStart(sync: SyncEntityInterface): Promise<number> {
```
vitalyiegorov: **P1 — an interrupted first run permanently truncates the backfill.**

While `forwardSyncedAt` is null, this falls back to `MIN(operatedAt)` over already-imported rows — a hard floor for the backward walk, not a resume cursor. `processSources` commits C2C first, and C2C is floored to 6 months; if the run then defers on the deadline (which the new deferral mechanism makes likely on a large first sync), the next run computes `earliestTransactionTime ≈ now−6mo` and uses it as the floor for **deposits, withdrawals, fiat and converts**. Everything older is never fetched, and once a run completes, `forwardSyncedAt` locks the window to 1 day. `resolveTransferWindowStart` below has the same defect and additionally derives the floor from **one arbitrary account's** earliest transaction while the transfers are fetched for all assets.

Fix: while backfill is incomplete, use `min(earliestTransactionTime, subYears(now, BINANCE_TRANSFER_LOOKBACK_YEARS))`, and track backfill completion with an explicit marker instead of inferring it from ledger contents. Note `getEarliestTransactionTimeByExternalSource` also counts soft-deleted rows, so delete-and-resync anchors on deleted data.

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1460 @@
...
+    private foldLockedEarnPositions(balanceByAsset: Map<string, number>, positions: BinanceLockedEarnPositionApiInterface[]): void {
+        for (const position of positions) {
+            const quantity = binanceMapper.parseBinanceAmount(position.amount);
+            if (isDefined(quantity) && quantity > 0) {
+                balanceByAsset.set(position.asset, (balanceByAsset.get(position.asset) ?? 0) + quantity);
+            }
+        }
+    }
+
+    private stripEarnPrefix(asset: string): string {
```
vitalyiegorov: **P1 — `stripEarnPrefix` corrupts any asset whose ticker starts with `LD` (e.g. LDO / Lido DAO).**

LDO is Binance-listed and Simple-Earn-eligible. A flexible LDO position folds into `balanceByAsset['O']`, which materializes a phantom `SPOT:O` account via `buildSpotAccounts`, while the real `SPOT:LDO` balance loses its entire Earn portion. The same prefix heuristic also excludes LDO from trade discovery at `collectBaseAssets` (:834) and `parseSelectedQuoteTradeSymbol` (:924), so LDO spot trades are never fetched either.

Fix: the flexible-position endpoint returns a plain `asset` (`BinanceEarnPositionApiSchema` has only `asset` + `totalAmount`), so drop the strip in `foldEarnPositions` entirely. If `LD*` ledger tokens really appear in `getUserAsset`, filter by an exact set built from the Earn positions actually returned, not a string prefix.

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1460 @@
...
+        return asset.startsWith(EARN_LD_PREFIX) ? asset.slice(EARN_LD_PREFIX.length) : asset;
+    }
+
+    private buildSpotAccounts(balanceByAsset: Map<string, number>): SyncAccountInterface[] {
+        const accounts: SyncAccountInterface[] = [];
+        for (const [asset, total] of balanceByAsset) {
+            if (total * MICRO_UNITS_PRECISION <= Number.MAX_SAFE_INTEGER) {
```
vitalyiegorov: **P1 — balances that overflow the 6-dp micro-unit range are anchored to exactly 0, indistinguishable from sold-off assets.**

Anything above ~9.007e9 units (routine for SHIB/PEPE/BONK/LUNC holders) is dropped from `getAccounts()` here (and in `parseBinanceAmount`/`computeTotalBalance`). The app's anchor loop then reads `balanceByExternalId.get(...) ?? 0` and — because `CRYPTO_SYNC` is bank-authoritative — forces the account to **zero** with no user-visible signal (the parked notice covers missing instruments, not overflow).

Fix: distinguish "absent from the exchange snapshot" (→ 0) from "present but unrepresentable" (→ leave the existing anchor untouched and surface as parked), by returning the unrepresentable assets explicitly from `getAccounts`.

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1460 @@
...
+            .filter(trade => trade.time >= startTimeMs && trade.time <= endTimeMs)
+            .map(trade => binanceMapper.mapTradeToTransfer(trade, symbol.baseAsset, symbol.quoteAsset))
+            .filter(isDefined);
+        const lastTrade = parsed.data.at(-1);
+        const nextFromId = parsed.data.length === TRADES_PER_SYMBOL_LIMIT && isDefined(lastTrade) ? lastTrade.id + 1 : null;
+
+        return this.success({ transfers, nextFromId });
+    }
+
+    private handleSymbolTradePageFailure(
```
vitalyiegorov, line 1066: **P1 — every `INVALID_RESPONSE` on a symbol page is swallowed as "unknown symbol", including transient HTTP 400s.**

`INVALID_RESPONSE` is what the base client returns for **any** 400 — including `-1021` recvWindow drift, which is realistic here because `serverTimeOffsetMs` is computed once per client, pinned to 0 on failure, and never retried, while a background run can span suspensions. Result: all trades for the symbol silently report zero, and `nextFromId: null` aborts paging so a mid-symbol failure truncates the rest of its history too.

Fix: swallow only Binance's actual invalid-symbol error (`code: -1121`) by parsing the body code (plumb it through `SyncError.originalError`), and propagate everything else.

packages/sync/src/binance/client/binance-signed.client.ts
```diff
@@ -0,0 +1,1460 @@
...
+        startTimeMs: number,
+        endTimeMs: number
+    ): Promise<SyncResultInterface<BinanceConvertFlowApiInterface[]>> {
+        return this.walkWindowsBackward(
+            { startTimeMs, endTimeMs, periodMs: CONVERT_MAX_PERIOD_MS, dormancyWindowCount: UNBOUNDED_DORMANCY_WINDOW_COUNT },
+            (windowStartMs, windowEndMs) => this.fetchConvertWindow(windowStartMs, windowEndMs)
+        );
+    }
+
+    private async fetchConvertWindow(
```
vitalyiegorov, line 763: **P1 — Convert flows are imported regardless of `orderStatus`.**

Fiat filters `status === 'Successful'` (:552) and C2C filters `orderStatus === 'COMPLETED'` (:1100), but there is no equivalent filter for converts — a FAILED/PROCESS convert imports phantom cross-asset transfers that never settled. The schema even requires `orderStatus`, so the data is right there.

Fix: filter `orderStatus === 'SUCCESS'` and add the negative scenario (`convert.test.ts` currently only ever uses the harness default `'SUCCESS'`).

packages/app/src/sync/service/abstract-polling-sync.service.ts
```diff
@@ -0,0 +1,338 @@
...
+            return false;
+        }
+
+        await syncRepository.recordError(syncToRetry.id, errorMessage);
+        await microPause(this.rateLimitMs);
+
+        return true;
+    }
+
+    private async disableFailedSyncs(enabledSyncs: SyncEntityInterface[], errorMessage: string): Promise<void> {
```
vitalyiegorov: **P1 — one non-retryable error on one source type disables every Binance sync row.**

`AppBinanceSyncService.isRetryableError` excludes `INVALID_RESPONSE`/`UNAUTHORIZED`, and every Zod `safeParse` failure or HTTP 400 in the client maps to `INVALID_RESPONSE`. Unlike Monobank (1–3 accounts), a Binance connection has one sync row **per asset**, so a single malformed row on one endpoint flips off dozens of accounts at once; recovery requires re-entering the token. The Zod schemas make this hair-triggered by requiring fields the client never reads (`BinanceTradeApiSchema.orderId/price/isMaker`, `BinanceConvertFlowApiSchema.quoteId`, `BinanceC2cOrderApiSchema.takerCommission`).

Fix: scope the disable to the failing sync row; drop the unread required fields; prefer row-level skip + warn over failing an entire page. Related (P2, gist): the deadline deferral also reuses `RATE_LIMITED`, so a genuine 429 now reports `completed:false, errorCount:0` and is silently absorbed — worth a dedicated deadline error code.

packages/app/src/sync/mapper/binance-transfer-input.mapper.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        const feeAccount = isDefined(transfer.feeAssetAccountId) ? await this.resolveAccount(transfer.feeAssetAccountId) : null;
+        const entries = [
+            this.buildEntry(fromAccount.id, TransactionEntryTypeEnum.CREDIT, transfer.fromAmount, transfer.externalId),
+            this.buildEntry(toAccount.id, TransactionEntryTypeEnum.DEBIT, transfer.toAmount, transfer.externalId)
+        ];
+        if (isPositiveNumber(transfer.feeAmount)) {
+            entries.push({
+                ...this.buildEntry(
+                    feeAccount?.id ?? fromAccount.id,
```
vitalyiegorov: **P1 — unresolvable fee assets are booked onto the wrong currency's account.**

`feeAccount?.id ?? fromAccount.id`: `commissionAsset` is frequently BNB (or any listed asset); when `resolveAccount` misses, a BNB-denominated amount is recorded as a FEE entry on the from-asset account (e.g. 0.0012 BNB booked as 0.0012 BTC), unscaled (`exchangeRate: 1`).

Fix: if the fee asset cannot be resolved, omit the fee entry and log it — never fall back to a different currency's account.

packages/sync/src/binance/client/binance-weight-throttle.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    recordHeaders(headers: Headers): void {
+        this.rawIpWeight = this.readWeight(headers, RAW_IP_WEIGHT_HEADER, this.rawIpWeight);
+        this.sapiIpWeight = this.readWeight(headers, SAPI_IP_WEIGHT_HEADER, this.sapiIpWeight);
+        this.sapiUidWeight = this.readWeight(headers, SAPI_UID_WEIGHT_HEADER, this.sapiUidWeight);
+    }
+
+    async waitIfNeeded(): Promise<void> {
+        if (this.shouldCoolDown()) {
+            if (Date.now() + WEIGHT_WINDOW_MS >= this.deadlineAtMs) {
```
vitalyiegorov, line 35: **P1 — the 25 s background budget is enforced only here, in the cool-down branch; the fetch loops never see it.**

A full-history batch issues 150+ sequential signed requests (deposit/withdrawal windows, convert/earn windows, per-symbol `myTrades` paging) and none of `walkWindowsBackward`, `fetchSymbolTrades`, `fetchNextOrderPage` or `signedRequest` checks `deadlineAtMs` — the only other consumer is `shouldStopProcessing()`, which runs **between sync rows**. A run of many fast requests blows through the budget and iOS kills it mid-row — which is exactly the state that triggers the first-run truncation defect in `resolveSourceWindowStart`.

Fix: check the deadline at the top of `signedRequest` (and between windows) and throw a **dedicated deadline error** so the run defers cleanly — reusing `SyncError.rateLimited` here is what makes real 429s unobservable (see gist P2 list).

packages/consolidation/src/query/repository/sql-factory/transfer-pair-ranked-candidate-base-sql.factory.ts
```diff
@@ -248,7 +270,23 @@ const TRANSFER_PAIR_RANKED_CANDIDATE_BASE_SQL = `
...
                         ELSE 0
-                    END as interbankHintedFeeAmountMatch
+                    END as interbankHintedFeeAmountMatch,
+                    CASE
+                        WHEN expense_entries.expenseInstrumentId != income_entries.incomeInstrumentId
+                            AND expense_entries.expenseEntryAmount > 0
+                            AND income_entries.incomeEntryAmount > 0
+                            AND (
+                                (income_entries.incomeAccountType = '${AccountTypeEnum.CRYPTO_SYNC}'
+                                    AND income_entries.incomeTransactionExternalId LIKE '%${P2P_ORDER_EXTERNAL_ID_MARKER}%'
```
vitalyiegorov: **P1 — the atomic path puts no currency or account-class constraint on the non-Binance leg.**

The whole gate is: different instruments, both amounts > 0, one leg `CRYPTO_SYNC` with `:c2c:`, the other merely `!= 'CRYPTO_SYNC'`. A `CASH`/`DEBT`/`STOCKS`/manual-`CRYPTO` expense within 1 h and 10 % of the triangulated cross-rate is auto-merged into a bogus "P2P fiat transfer" — e.g. a 100 EUR cash expense at 12:30 + a ~117 USDT P2P buy at 12:45 paid from an account the user never imported → the real cash expense disappears from statistics. The authoritative path is immune (it pins `bank.instrumentId = p2p.quotedInstrumentId`); the atomic fallback runs precisely when that quote is missing.

Fix: carry `quoted_instrument_id` into this CTE and require the counter-leg's `account.instrument_id` to match when present; when absent, at minimum require `instrument.type = FIAT` and exclude non-spendable account types.

vitalyiegorov (commented):
Round 3 — full deep review of the current branch state (engine, consolidation SQL, app UI, conventions, tests; SQL claims verified with `EXPLAIN QUERY PLAN` against the real migrated schema). Full analysis with the complete P2 list and coverage map: https://gist.github.com/vitalyiegorov/57d956c10fffb6c5bcd0ef44712ef990

**Rounds 1–2 were substantively addressed** — 15/19 findings verified fixed on `9249e65f0` with code evidence. Partial: the BGTask deadline reaches only the throttle cool-down branch (not the window walkers), and the foreground still runs with `deadline = Infinity` so a 60 s cool-down can sleep in the interactive runtime. Still open (minor): `paceRequestRate` never re-checks after its delay; `getTransfers(accountId, …)` still ignores `accountId`; the "(temporary)" `max-statements` justification; inline `SetupStep` type.

This round found **two P0 data-integrity defects on mainline paths** (inline: frozen monthly Earn buckets; FEE entries leaking into the P2P matcher) plus P1s in two clusters — **silent data truncation** (deposit paging, first-run floor, `LD*` prefix corruption, overflow→0 anchoring, swallowed 400s, unfiltered converts, whole-provider disable, budget unenforced in fetch loops, currency-unconstrained atomic match) and **late-rename regressions** (`@Log` → inline `getLogger` in three services, raw `useLiveQuery`, dropped `isSubmitting`/sticky instrument, three translations blanked in all four locales, AGENTS.md reverted to stale content). All inline.

**P2 highlights** (details in the gist): a genuine 429 is now indistinguishable from the soft deadline (`isDeadlineDeferral` keys on `RATE_LIMITED` — real rate-limits report `completed:false, errorCount:0`; introduce a dedicated deadline error code); run-scoped caches never hit because every entry point derives `endTimeMs` from a fresh `Date.now()` (~2× request volume); the matcher compares historical trades against **today's** spot rate while `historical_exchange_rates` + `getExchangeRateWithHistoricalFallbackSql` sit unused; `AUTO_P2P_FIAT_TRANSFER` is dead code and `'p2p-fiat'` matchType is unreachable; `collectExpenseCombinations` is unbounded below (C(n,≤3), no yield); `existing-transfer-bridge` lacks the consolidation-type gate its siblings have and can grab a P2P canonical next run; locked-Earn rewards are never imported while locked principal folds into the anchored balance (permanent drift); P2P `takerCommission` parsed then discarded; earn month keys are UTC vs local display; boundary-ms windows double-sum earn rewards; transfer legs share one `externalId`.

**Tests to land with the fixes**: 429-during-sync leaves the sync enabled (the new retryable classification is unprotected); `sync-provider-registry.test.ts` never tests Binance; Simple-Earn month-boundary case (doubles as the P0-1 repro — two syncs in one month asserting the bucket total); Convert non-SUCCESS status; amount assertions on Convert/deposit (existence-only today — a halved amount passes); fee-attribution `accountId`/amount; resume-after-interrupt. The P2P matching/repair suites, credential rotation, and the unconsolidate round-trip are excellent as-is.

**Worth checking against a live key before merge** (both fail loudly if wrong): `type: 'ALL'` on the flexible rewards endpoint, and `/api/v3/myTrades` with `startTime` but no `endTime` (a 24 h default range would silently fetch zero trades).

Recommendation: fix the two P0s, the truncation cluster, and the rename regressions before merge; the rest can join #562.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/app/src/account/component/update-liability-account/update-liability-account.tsx
```diff
@@ -39,7 +38,7 @@ export const UpdateLiabilityAccount = ({ account }: Props) => {
         isActive: account.isActive
     };
 
-    const { control, handleSubmit, instrument, isSubmitting } = useAccountForm(
+    const { control, handleSubmit, instrument } = useAccountForm(
```
vitalyiegorov: **P1 — two regressions bundled into this rename edit.**

1. `isSubmitting` was dropped from the `useAccountForm(...)` destructure and is no longer passed to `<UpdateAccountScreen>` — the Update button loses its loading/disabled state during submit. `create-liability-account`, `create-debt-account` and `update-debt-account` all still thread it.
2. `useStickyDefinedValue(instrument)` was replaced with a direct `isDefined(instrument)` check — this screen alone now flashes `<EmptyScreen />` during instrument refetches, while its three siblings keep the sticky value.

Restore both.

packages/app/src/sync/hook/use-account-sync.hook.ts
```diff
@@ -0,0 +1,16 @@
+import { useLiveQuery } from 'drizzle-orm/expo-sqlite';
```
vitalyiegorov: **P1 — raw `useLiveQuery` instead of `useDatabaseLiveQuery` (rule 60).**

This is the only raw `drizzle-orm/expo-sqlite` `useLiveQuery` call site in the app, and it's a regression from the deleted `use-account-bank-sync.hook.ts`, which correctly used `useDatabaseLiveQuery(bankSyncRepository.findByAccountId(accountId), [accountId])`. `AccountSyncCard` may not refresh sync status/toggle/token after service writes that invalidate via `@InvalidateDatabaseLiveQuery`.

packages/app/src/i18n/hook/use-format-instrument-amount.hook.ts
```diff
@@ -0,0 +1,15 @@
+import { InstrumentTypeEnum } from '@budgie/contracts';
+
+import { MICRO_UNIT_DECIMAL_PLACES } from '../../@generic/constant/micro-unit-decimal-places.constant';
+import { useSettingsContext } from '../../settings/context/settings.context';
+
+import { useFormatDigits } from './use-format-digits.hook';
+
+export const useFormatInstrumentAmount = () => {
```
vitalyiegorov, line 8: **P1 — the crypto branch of this hook is unreachable, while a second copy of the same formatting diverges elsewhere.**

Sole consumer is `transaction-fee-pill.tsx`, and both of *its* call sites (`transfer-quick-form.tsx:186`, `simple-quick-form-fee-pill.tsx:15`) omit `instrumentType`, so it is always `null` — the crypto fee formatting described in the PR summary ships dead. Meanwhile `crypto-account-card.tsx:20-23` re-implements the identical `formatDigits` + `"${amount} ${symbol}"` logic inline.

Fix: thread `instrumentType` through the fee-pill call sites and reuse this hook in `crypto-account-card` — or, if fee pills are intentionally fiat-only for now, drop the branch/prop (the hook is then borderline single-consumer under rule 52).

packages/app/src/app/(main)/create-account/index.tsx
```diff
@@ -47,24 +47,30 @@ export default function Index() {
...
                 bankProvider={ExternalSourceEnum.PRIVATBANK}
             />
-            <CreateBankSyncCard
+            <CreateSyncCard
                 description={t`Import accounts and transactions from Erste Bank PDF statement`}
                 title={t`Erste Bank`}
                 route={ersteRoute}
                 bankProvider={ExternalSourceEnum.ERSTE}
             />
+            <CreateSyncCard
```
vitalyiegorov, line 69: **P1 — the Binance card now sits under a `<Trans>Bank Sync</Trans>` section header.**

Binance is a crypto exchange, not a bank — this is the only remaining "Bank Sync" literal in app source and it undercuts the PR's own bank-sync→sync rename rationale (the component itself was correctly renamed `CreateSyncCard` in this same diff). Retitle and run `yarn i18n:sync`. Minor (P2): the route here is a raw template literal while the three siblings use `as const` typed route objects.

packages/app/src/sync/service/monobank-sync.service.ts
```diff
@@ -1,134 +1,106 @@
...
-    MccCategoryLookupInterface,
-    TransactionEntityInterface
-} from '@budgie/contracts';
+import type { MccCategoryLookupInterface, SyncEntityInterface, TransactionEntityInterface } from '@budgie/contracts';
+import type { SyncAccountInterface, SyncBatchResultInterface } from '@budgie/sync';
 
-class AppMonobankSyncService {
-    private static readonly BACKGROUND_TASK_MINIMUM_INTERVAL_MINUTES = 15;
-    private static readonly FORWARD_SYNC_STALE_THRESHOLD_MS = 120000;
+const logger = getLogger('AppMonobankSyncService');
```
vitalyiegorov: **P1 — logging regression (rule 32): module-scope `getLogger` + inline `logger.log` inside classes.**

At merge-base this file used `@Log` decorators exclusively (18 occurrences, zero `getLogger`); the rewrite replaced lifecycle decoration with ~16 ad-hoc inline calls. The new `binance-sync.service.ts` (`getLogger` at :51) has the same shape, and `privatbank-category-matcher.service.ts` was additionally demoted from a `@Log`-decorated class to free functions (`privatbankCategoryMatcherMatch` + a module-level single-consumer helper), losing enter/done/throw coverage entirely (rules 22/32/43).

Fix: restore the class shapes and split the logged phases into granular private methods decorated with `@Log`.

AGENTS.md
```diff
@@ -1,6 +1,6 @@
 # CLAUDE.md
 
-Budgie is an offline-first mobile expenses tracker. The production monorepo contains app, contracts, ai, landing, bank-sync, budget, consolidation, and logger packages.
+Budgie is an offline-first mobile expenses tracker. Monorepo with 5 packages: app (React Native), contracts (shared types), ai (AI/LLM services), landing (Next.js), and sync (bank integrations).
```
vitalyiegorov: **P1 — AGENTS.md was reverted to stale content; looks like a wrong-side merge pick.**

The branch version describes Prettier/ESLint while `package.json` (unchanged by this PR) runs oxfmt/oxlint; drops the `budget`/`consolidation`/`logger` commit scopes (this PR itself edits `packages/consolidation`); deletes the Agent Orchestration section; downgrades "Expo 56"→"Expo 54" and "yarn 4.17"→"4.12"; and points rule 61's grandfather list at `eslint.config.mjs` instead of `.oxlintrc.json`.

Fix: re-base this file on main's content and apply only the `bank-sync`→`sync` substitutions. Also in this cluster: three msgids (`"Add transactions to see how your spending breaks down by category."`, `"Bank Sync"`, `"No categories in this period"`) had their translations blanked in **all four** locales' `.po` files — the compiled catalogs now ship English fallback for de/es/fr/uk.

PR #613, opened 2026-07-24
chore(consolidation): cover casing variants of PrivatBank reversal title prefixes

vitalyiegorov:
## Summary

SQLite's `UPPER()` is ASCII-only, so it never folds Cyrillic casing. `rawNormTitle` (`UPPER(TRIM(title))`) therefore leaves Cyrillic title casing untouched, and each PrivatBank rejected-payment reversal title casing variant must be registered as its own `LIKE` literal — the same pattern already used for `AUTO_TITLE_PREFIXES` (both `'ПОВЕРНЕННЯ КОШТІВ, '` and `'Повернення коштів, '` are listed).

`REJECTED_PAYMENT_PRINCIPAL_TITLE_PREFIX` and `REJECTED_PAYMENT_FEE_TITLE_PREFIX` were title-case-only string literals used via `LIKE '<prefix>%'` in two places each (the `confidenceBucket` CASE and the `matchType` CASE inside `buildCompatiblePairsSql`). If PrivatBank ever emits an ALL-CAPS variant of these prefixes, those reversals would silently stop auto-matching.

## Changes

- Converted both constants into `readonly` arrays of casing variants (title-case + ALL-CAPS): `REJECTED_PAYMENT_PRINCIPAL_TITLE_PREFIXES`, `REJECTED_PAYMENT_FEE_TITLE_PREFIXES`.
- Added `buildPrefixLikeSql(column, prefixes)`, a small helper producing an OR'd `LIKE` condition (e.g. `(col LIKE 'a%' OR col LIKE 'b%')`), used in both the `confidenceBucket` and `matchType` CASE expressions, keeping the two aligned.
- Added a regression test in `tests/consolidation-tests/src/scenarios/refund-pair-rejected-payment.test.ts` asserting an ALL-CAPS-prefixed refund title auto-consolidates the same as the title-case one, using a new `REJECTED_PAYMENT_PRINCIPAL_TITLE_ALL_CAPS` fixture constant.

Fixes #608

## Test plan

- [x] Added the regression test first, confirmed it failed against the unfixed factory (`consolidated` was `0` instead of `1`)
- [x] Implemented the fix, confirmed the same test passes
- [x] `yarn workspace @budgie-at/consolidation-tests test` — 24 files / 62 tests passed
- [x] `yarn workspace @budgie/consolidation ts` — clean
- [x] `yarn workspace @budgie-at/consolidation-tests ts` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #612, opened 2026-07-24
fix(consolidation): fee-return refund rejected when FEE entry exceeds primary CREDIT amount

vitalyiegorov:
## Root cause

`REFUND_AUTO_CANDIDATES_SQL` bounded every candidate group with `HAVING SUM(refundAmount) <= expenseAmount`, always comparing against the primary **CREDIT** amount.

For the `AUTO_REFUND_REJECTED_PAYMENT_FEE_TITLE` bucket, the refund is matched against the expense **FEE** entry (`inc.amount = exp.feeAmount` in `buildCompatiblePairsSql`). The bank-sync mapper (`packages/bank-sync/src/core/mapper/map-bank-transaction-to-create-input.mapper.ts`) sets `mainAmount = amount - feeAmount`, so whenever a fee exceeds 50% of the charged total, `feeAmount > expenseAmount` and the `HAVING` bound silently rejected a perfectly valid fee-return refund (false negative).

## Fix

- Carry `exp.feeAmount AS feeAmount` through the `compatible_pairs` CTE (propagates via `SELECT *` in the ranked CTEs).
- Bound the fee bucket against the FEE amount: `HAVING SUM(refundAmount) <= CASE WHEN confidenceBucket = 'AUTO_REFUND_REJECTED_PAYMENT_FEE_TITLE' THEN feeAmount ELSE expenseAmount END`.
- `feeAmount` is constant within each `(expenseTxId, accountId)` group, so it is added to `GROUP BY` (SQLite-safe, consistent with the existing `expenseAmount` handling).
- `REFUND_REVIEW_CANDIDATES_SQL` is left unchanged: its `WHERE` only admits `REVIEW_REFUND_PREFIX_TITLE_MCC` and `AUTO_REFUND_LOCALIZED_REFUND_TITLE`, so the fee bucket can never reach its `HAVING`. No window/rank logic needs `feeAmount`.

## Tests

- New high-fee-ratio regression scenario in `tests/consolidation-tests/src/scenarios/refund-pair-rejected-payment.test.ts` (primary credit 100, fee 150 — fee exceeds the primary credit). Confirmed it fails on `origin/main` (`consolidated=0`) and passes after the fix. Extends the shared `rejected-payment-fixture.ts` rather than duplicating.
- Full suite green: `yarn workspace @budgie-at/consolidation-tests test` -> 62 passed (incl. `refund-pair-rejected-payment-statistics.test.ts`).
[...]

PR #614, opened 2026-07-24
fix(consolidation): route overflow shared-expense refunds to review instead of dropping the group

### LINKED ISSUE #604: fix(consolidation): shared-expense refund group silently dropped when second refund bypasses mutual-best gate

When two refund incomes compete for ONE exact-amount expense and the losing refund's only compatible candidate is that same expense, the `refundCandidateCount = 1` acceptance branch fires unconditionally and bypasses the mutual-best uniqueness gate. Both refunds then land in the same GROUP BY group (same bucket/matchType/expenseTxId/accountId/expenseAmount) and `HAVING SUM(refundAmount) <= expenseAmount` silently drops the WHOLE group — the legitimate winner gets neither auto-consolidated nor surfaced for review.

Where: `packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts` — acceptance branch A (`refundCandidateCount = 1`) and the HAVING bound in `REFUND_AUTO_CANDIDATES_SQL`.

Expected: accept the mutual-best winner, route the loser to manual review; never zero out the group silently.

Needs a regression test in `tests/consolidation-tests` (two refunds, one shared exact-amount expense, loser has no other candidates).

vitalyiegorov:
## Root cause

In `REFUND_AUTO_CANDIDATES_SQL` the acceptance branch `refundCandidateCount = 1` fires **per refund** with no expense-side coordination. When two refund incomes each have the same exact-amount expense as their *only* candidate, both are accepted, both land in the same auto `GROUP BY` group `(confidenceBucket, matchType, expenseTxId, accountId, expenseAmount)`, and the `HAVING SUM(refundAmount) <= expenseAmount` guard silently drops the **whole** group. The legitimate winner is then neither auto-consolidated nor surfaced for manual review — it just vanishes.

## Fix

`packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts`:

- **Running total window** — `ranked_candidates` now computes `expenseRunningRefundTotal`: `SUM(CASE WHEN refundRank = 1 THEN refundAmount ELSE 0 END) OVER (PARTITION BY expenseTxId, accountId, confidenceBucket, matchType, expenseAmount ORDER BY expenseRank ROWS UNBOUNDED PRECEDING)`. Only primary (`refundRank = 1`) refunds — the ones actually competing to consolidate onto that expense — accumulate against its capacity, ordered by `expenseRank` so the best-ranked refund wins the room first.
- **Auto path** — additionally requires `expenseRunningRefundTotal <= expenseAmount`, so only the best-ranked refund(s) that fit are auto-accepted.
- **Review path** — gains a fallback branch: refunds that pass an auto-acceptance gate but overflow the running-total bound (`expenseRunningRefundTotal > expenseAmount`) now surface for manual review instead of disappearing.
- The shared acceptance predicate is extracted into `buildAutoAcceptanceGateSql()` consumed by both paths (no duplicated SQL), and the defensive `HAVING` invariant is kept.

Legitimate multiple partial refunds (e.g. 30 + 30 against a 100 expense) still auto-consolidate together, because their cumulative running total stays within the expense amount.

## Tests

- New scenario `tests/consolidation-tests/src/scenarios/refund-pair-shared-expense.test.ts`: two refund incomes competing for one exact-amount expense where the loser has no other candidate. Asserts the winner appears as the sole auto candidate and auto-consolidates, while the loser surfaces as a review candidate (not silently dropped) and stays unconsolidated. Confirmed failing on `main` before the fix.
- Full suite green: `@budgie-at/consolidation-tests` 62/62 pass; `@budgie/consolidation` and `@budgie-at/consolidation-tests` typecheck clean; consolidation lint clean.

Fixes #604

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved refund matching for expenses with multiple refunds.
  - Automatically consolidates eligible refunds only while the combined refund amount remains within the expense total.
  - Routes refunds that exceed the expense amount to review instead of auto-consolidating them.

- **Tests**
  - Added coverage validating shared-expense refund handling, including successful consolidation and review routing for overflow refunds.

### DIFF
--- a/packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts
+++ b/packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts
@@ -206,9 +206,32 @@ const buildRankedCandidatesSql = (): string => `
                         AND expenseRank = 1
                     THEN 1 ELSE 0
                 END
-            ) OVER (PARTITION BY refundTxId) AS localizedExactAmountMatchCount
+            ) OVER (PARTITION BY refundTxId) AS localizedExactAmountMatchCount,
+            SUM(CASE WHEN refundRank = 1 THEN refundAmount ELSE 0 END) OVER (
+                PARTITION BY expenseTxId, accountId, confidenceBucket, matchType, expenseAmount
+                ORDER BY expenseRank
+                ROWS UNBOUNDED PRECEDING
+            ) AS expenseRunningRefundTotal
         FROM ranked_pairs
     )
+`;
+
+const buildAutoAcceptanceGateSql = (): string => `
+        confidenceBucket IN (
+            'AUTO_REFUND_EXACT_TITLE',
+            'AUTO_REFUND_LOCALIZED_REFUND_TITLE',
+            'AUTO_REFUND_REJECTED_PAYMENT_PRINCIPAL_TITLE',
+            'AUTO_REFUND_REJECTED_PAYMENT_FEE_TITLE'
+        )
+        AND (
+            refundCandidateCount = 1
+            OR (
+                confidenceBucket = 'AUTO_REFUND_LOCALIZED_REFUND_TITLE'
+                AND expenseAmount = refundAmount
+                AND expenseRank = 1
+                AND localizedExactAmountMatchCount = 1
+            )
+        )
 `;
 
 const buildRankedCandidateSql = (scope: ConsolidationScanScopeInterface | null): string => {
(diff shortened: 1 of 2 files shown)


vitalyiegorov:
Closing temporarily to relieve CI pressure — we're landing this refund-bug series strictly one PR at a time, starting with #611.

The branch is **not** deleted and no work is lost; this PR will be reopened (and rebased onto the new `main`) when its turn comes. Sequence: #611 → #613 → #612 → #614 → #615.

PR #615, opened 2026-07-24
fix(consolidation): cross-bucket priority can hide a unique localized exact-amount refund match

### LINKED ISSUE #605: fix(consolidation): cross-bucket priority can hide a unique localized exact-amount refund match

Refund candidate selection keeps only the `refundRank = 1` row per refund income. Bucket priority (EXACT_TITLE > LOCALIZED > ...) outranks the amount-mismatch sort key, so if a refund's rank-1 row is an EXACT_TITLE candidate to a DIFFERENT expense while its unique localized exact-amount match sits at rank 2, the localized row never reaches the auto or review WHERE clauses. If the rank-1 row itself then fails its own acceptance (`refundCandidateCount > 1`) and has no review fallback for its bucket, the refund is silently dropped — neither consolidated nor reviewable.

Where: `packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts` — `refundRank` window ordering (`buildBucketPriorityOrderSql` + `buildLocalizedAmountMismatchOrderSql`) and the rank-1 filtering in `REFUND_AUTO_CANDIDATES_SQL` / `REFUND_REVIEW_CANDIDATES_SQL`.

Needs a regression test constructing the cross-bucket collision before changing ranking semantics.

vitalyiegorov:
## Root cause

Both `REFUND_AUTO_CANDIDATES_SQL` and `REFUND_REVIEW_CANDIDATES_SQL` in `refund-ranked-candidate-sql.factory.ts` filter to `refundRank = 1` per refund income. The rank ordering is bucket priority first (`EXACT_TITLE` = 1 > `LOCALIZED` = 2 > ...), then localized amount-mismatch, then `timeDiff`.

So when a refund's rank-1 row is an `AUTO_REFUND_EXACT_TITLE` candidate against expense E1, while its unique localized exact-amount match (`AUTO_REFUND_LOCALIZED_REFUND_TITLE` against E2, with `expenseAmount = refundAmount`, mutual-best, `expenseRank = 1`) sits at rank 2, the localized row never reaches either WHERE clause. If the rank-1 `EXACT_TITLE` row then fails auto-acceptance (`refundCandidateCount > 1`, and the localized mutual-best branch does not apply to its bucket), it has no review fallback (review only covered `AUTO_REFUND_LOCALIZED_REFUND_TITLE`). The refund is silently dropped: neither consolidated nor reviewable.

## Chosen design — option (a)

Allow the localized mutual-best acceptance branch (`confidenceBucket = 'AUTO_REFUND_LOCALIZED_REFUND_TITLE' AND expenseAmount = refundAmount AND expenseRank = 1 AND localizedExactAmountMatchCount = 1`) to accept its row **regardless of `refundRank`**, while the other buckets still require `refundRank = 1 AND refundCandidateCount = 1`.

Why (a) over generalizing the review fallback (b):

- The rank-1 row in this collision is the *contested* exact-title candidate against the wrong expense (E1). Option (b) would surface *that* weaker pairing for manual review while still hiding the genuinely strong, unique localized exact-amount match (E2). Option (a) recovers the correct match (E2) automatically.
- **Provably free of double consolidation.** When the localized branch fires, `refundCandidateCount > 1`, so the `(refundRank = 1 AND refundCandidateCount = 1)` clause is false for every row of that refund; and `localizedExactAmountMatchCount = 1` guarantees exactly one qualifying row per refund. A refund is therefore never consolidated against two expenses. Because the strong match is recovered by auto (not review), there is no double-surfacing across the auto/review queries.
- The exact-title / rejected-payment buckets are unchanged (they were previously `refundRank = 1 AND refundCandidateCount = 1` in effect, and remain so). The only behavioral change is that a unique localized exact-amount mutual-best row at `refundRank > 1` is now accepted — precisely the reported gap.

## Tests

Added `tests/consolidation-tests/src/scenarios/refund-pair-cross-bucket-localized-exact-amount.test.ts`, constructing the cross-bucket collision via `testSeedService`: refund R matches E1 by exact-title (E1 also claimed by another refund R2) and matches E2 by a unique localized exact-amount mutual-best. The test was confirmed to fail on the pre-fix code (R vanishes from both auto and review candidates) and passes after the fix (R auto-consolidates against E2, and never surfaces paired with E1 nor in review).

- `yarn workspace @budgie-at/consolidation-tests test` — 62 passed (25 files)
- `yarn workspace @budgie/consolidation ts` — pass
- `yarn workspace @budgie-at/consolidation-tests ts` — pass
- `yarn workspace @budgie/consolidation lint` — pass

Fixes #605

### DIFF
--- a/packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts
+++ b/packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts
@@ -238,15 +238,14 @@ export const REFUND_AUTO_CANDIDATES_SQL = (scope: ConsolidationScanScopeInterfac
         SUM(refundAmount) AS refundsTotal,
         GROUP_CONCAT(refundTxId, ',' ORDER BY refundTxId) AS refundIncomeTransactionIds
     FROM (${buildRankedCandidateSql(scope)})
-    WHERE refundRank = 1
-        AND confidenceBucket IN (
+    WHERE confidenceBucket IN (
             'AUTO_REFUND_EXACT_TITLE',
             'AUTO_REFUND_LOCALIZED_REFUND_TITLE',
             'AUTO_REFUND_REJECTED_PAYMENT_PRINCIPAL_TITLE',
             'AUTO_REFUND_REJECTED_PAYMENT_FEE_TITLE'
         )
         AND (
-            refundCandidateCount = 1
+            (refundRank = 1 AND refundCandidateCount = 1)
             OR (
                 confidenceBucket = 'AUTO_REFUND_LOCALIZED_REFUND_TITLE'
                 AND expenseAmount = refundAmount
--- a/tests/consolidation-tests/src/scenarios/refund-pair-cross-bucket-localized-exact-amount.test.ts
+++ b/tests/consolidation-tests/src/scenarios/refund-pair-cross-bucket-localized-exact-amount.test.ts
@@ -1,0 +1,61 @@
+import { PRECISION, TransactionConsolidationTypeEnum } from '@budgie/contracts';
+import { describe, expect, it } from 'vitest';
+
+import { runConsolidation } from '../harness/run-consolidation';
+import { refundPairRepository, testQueryService, testSeedService } from '../harness/test-context';
+
+const REFUND_TITLE = 'Скасування. Netflix';
+const LOCALIZED_EXPENSE_TITLE = 'Netflix';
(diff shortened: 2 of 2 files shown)


vitalyiegorov:
Closing temporarily to relieve CI pressure — we're landing this refund-bug series strictly one PR at a time, starting with #611.

The branch is **not** deleted and no work is lost; this PR will be reopened (and rebased onto the new `main`) when its turn comes. Sequence: #611 → #613 → #612 → #614 → #615.

PR #617, opened 2026-07-25
ci(app): move Android release lint off the store build and trim it to shipping ABIs

vitalyiegorov:
## Why

The `production` Android build profile carried no lint exclusions. The `e2e` profile in the same file did:

```jsonc
// packages/app/eas.json — before this PR
"e2e": {
    "android": {
        "gradleCommand": ":app:assembleRelease -x lint -x lintVitalAnalyzeRelease -x lintAnalyzeRelease"
    }
},
"production": {
    "autoIncrement": true,
    "environment": "production",
    "env": { "APP_VARIANT": "production" },
    "channel": "production"
    // no "android" key at all -> default gradleCommand
}
```

The Android Gradle plugin wires **`lintVitalRelease`** into release assembly and bundling, so the default `:app:bundleRelease` runs full Android Lint inline inside one Gradle step.

**Measured on the sibling repository** (`vitalyiegorov/suuudokuuu`, same Expo/React Native monorepo stack and the same self-hosted Apple-silicon builder), not on this repository: the identical defect cost **more than 90 minutes** in a single Gradle step, against **12.1 minutes** for the byte-identical build with those three exclusions applied. Only `lintVital` is in the release task graph — plain `lint` and `lintAnalyzeRelease` are not, so excluding only the non-vital variants saves ~0s. The vital task is the one that has to move off the critical path.

Second defect, confirmed by `expo prebuild` in this repository: nothing in `eas.json`, `app.config.js` or the Expo config restricted Android ABIs, so a store build compiled native C++ four times.

## What this repository does and does not have

**This repository has no Android build or publish job.** `native-publish.yml` builds and submits iOS only; `pr.yml` has `build-ios-e2e-app` and `e2e-ios` and no Android equivalent. Production Android builds happen out of band, through `eas build --platform android --profile production`, and hit the defect above in full.

Consequences for this port:

- There is **no `needs:` edge to add**, because there is no Android submission job to gate. The new lint gate runs as its own job in the store publish workflow, and its comment records that any Android build-and-submit job added to this workflow **must** declare `needs: android-release-lint`. Please keep that comment intact.
- The 64-bit ABI guard runs against the **generated production Gradle project** rather than a packaged `.aab`, because no job in this repository produces an Android artifact to inspect.

## Changes

| Change | Expected saving |
|---|---|
| `production.android.gradleCommand` gets the same three lint exclusions the `e2e` profile already uses, on `:app:bundleRelease` (the production profile has no `distribution` override, so it is a store AAB build) | the bulk of the 90min -> 12min delta measured on the sibling repository |
| New `android-release-lint` job on hosted `ubuntu-24.04` runs `:app:lintVitalRelease` against the production Expo config, and uploads the HTML/XML lint report on failure | keeps the fatal-severity check, off the macOS fleet, which has zero spare capacity |
| Production variant restricted to `armeabi-v7a` + `arm64-v8a` via `expo-build-properties` `buildArchs` | ~110s of CMake compile on the sibling's per-ABI measurements (x86 67s, x86_64 43s), plus per-ABI configure, `mergeReleaseNativeLibs`, strip and packaging, and two fewer copies of every `.so` in the artifact |
| `ndk;27.0.12077973` and `cmake;3.22.1` added to the lint job's SDK packages, with a warning when either NDK is absent | up to ~45s of mid-build NDK download (see below) |
| `Verify production Android ABIs` step before lint | ~0s, guards the ABI change |

### The expo-updates NDK fallback applies here too

`node_modules/expo-updates/android/build.gradle` (57.0.6) declares `externalNativeBuild { cmake { ... } }` but, unlike `expo-modules-core`, never reads `rootProject.ext.ndkVersion`:

```groovy
// expo-modules-core/android/build.gradle
if (rootProject.hasProperty("ndkVersion")) {
  ndkVersion rootProject.ext.ndkVersion
}

// expo-updates/android/build.gradle — no equivalent
```

`ExpoRootProjectPlugin.kt` sets `ndkVersion` to `27.1.12297006` from the React Native version catalog (`node_modules/react-native/gradle/libs.versions.toml`: `ndkVersion = "27.1.12297006"`, `agp = "8.12.0"`, `compileSdk = "36"`), so AGP falls back to its own default `27.0.12077973` and downloads it during configuration. The `27.0.12077973` value is inherited from the sibling repository's Gradle log, which showed a 45.0s `Install NDK (Side by side) 27.0.12077973` inside `> Configure project :expo-updates`; it was not independently measured here. The lint job installs both NDKs up front. **When an Android build job is added to this repository, its `setup-android` step needs the same two `ndk;` entries plus `cmake;3.22.1`.**

## ABI product impact — flagged for human review

Dropping `x86` and `x86_64` from the production build means:

- **Android emulators cannot run the store build.** Only the production variant is affected; development, preview and E2E builds are untouched.
- **Intel Chromebooks** fall back to ARC native-bridge translation for the ARM libraries instead of running native x86_64 code. Apps generally still work, with a startup and CPU cost. ARM Chromebooks are unaffected.
- **Play Games on PC** requires x86_64 and would no longer be served this build.

Nothing in this repository indicates Chromebook, ChromeOS, emulator or Play Games on PC distribution: a full-tree search for `chromebook`, `chrome os`, `chromeos`, `play games` and `buildArchs` found no product references outside iOS `EXCLUDED_ARCHS` lines and `uname -m` host assertions. **If any of the above matters for budgie, veto this change** — it is isolated in commit `5f5b77c` (`app.config.js` plus the verify step and NDK entries) and can be reverted without losing the lint fix in `1ca897d`.

## Verification

Run locally against this branch:

- `expo prebuild --platform android` per variant, reading `android/gradle.properties`:

  | `APP_VARIANT` | `reactNativeArchitectures` |
  |---|---|
  | `production` | `armeabi-v7a,arm64-v8a` |
  | `e2e` | `armeabi-v7a,arm64-v8a,x86,x86_64` |
  | `preview` | `armeabi-v7a,arm64-v8a,x86,x86_64` |
  | `development` | `armeabi-v7a,arm64-v8a,x86,x86_64` |
  | unset | `armeabi-v7a,arm64-v8a,x86,x86_64` |

- `actionlint` on `native-publish.yml`: no new findings. The only report is the pre-existing `macos-tartelet` custom self-hosted label on line 15, which every workflow in the repository already produces.
- YAML parses; both jobs present; no `needs:` edges, so the graph is trivially acyclic.
- `bash -n` clean on all six `run:` blocks in the new job.
- The 64-bit guard was executed against good and bad inputs under `bash -e`:

  | `reactNativeArchitectures` | exit |
  |---|---|
  | `armeabi-v7a,arm64-v8a` | 0 |
  | `armeabi-v7a,arm64-v8a,x86,x86_64` | 0 |
  | `arm64-v8a` | 0 |
  | `armeabi-v7a` | 1 |
  | `armeabi-v7a,x86` | 1 |
  | property absent | 1 |
  | `arm64-v8a-fake` | 1 |

- The NDK warning loop was executed with a fake `ANDROID_HOME`: it warns for the missing version only, and does not abort the step.
- `yarn ts` 13/13 tasks pass, `yarn lint` clean, `yarn deadcode` reports only pre-existing configuration hints, `yarn cpd` finds 0 clones.
- Both commit subjects pass the repository commitlint config.

**Not verified:** the `android-release-lint` job has never executed, because `native-publish.yml` is `workflow_dispatch` only and no Android CI exists here to borrow evidence from. In particular, whether `:app:lintVitalRelease` completes on `ubuntu-24.04` within the 60-minute timeout, and whether it needs anything beyond the installed SDK packages, is untested. Dispatch this workflow once before relying on the gate.

## Not changed

Signing, credential creation, artifact upload, submission, and version/build-number logic are untouched. The iOS job is byte-identical; no `needs:` was added to it, so iOS submission is not coupled to the Android lint gate. No other workflow was modified.

Separate pre-existing finding, deliberately left alone as out of scope: the iOS publish job writes `packages/app/asc_api_key.p8` into the workspace and has no cleanup step, unlike the sibling repository, which removes it with `rm -f packages/app/asc_api_key.p8` under `if: always()`. To be accurate about severity: this is **not** a host-level key leak. The runners are ephemeral Tart VMs, destroyed at job end, and there are no `.p8` files on the host, so the key never outlives its own VM. It is worth closing as defense-in-depth (parity with the sibling, in-VM blast radius, and the house rule against persisting private keys), which is done separately in #618 rather than here.

## Summary by CodeRabbit

* **New Features**
  * Added automated Android production lint checks to catch issues before release.
  * Production Android builds now target both 32-bit and 64-bit ARM architectures.
  * Production Android builds now generate a release bundle as part of the build process.
* **Bug Fixes**
  * Added safeguards to ensure 64-bit ARM (arm64) is included in production Android builds, failing the release step if it’s missing.


vitalyiegorov:
Rebased onto `main` (`e846fd96`) after #618 merged.

**What conflicted.** Only `.github/workflows/native-publish.yml`, and only because both changes appended to the same place — #618 added the `Remove App Store Connect key` step after `Submit iOS app`, and this branch added the `android-release-lint` job after the same line. Resolved by keeping **both** sides in the correct order: the cleanup step stays the final step of the iOS `build-and-publish` job, and the Android lint job follows it as a separate job. `packages/app/eas.json` and `packages/app/app.config.js` did not conflict.

**#618's cleanup step is preserved, unmodified.** Verified three ways:

- The workflow hunk is `@@ -81,3 +81,97 @@` — pure addition. The three `Remove App Store Connect key` lines appear as *context*, not deletions.
- The whole PR diff contains exactly one deleted line, `-            "channel": "production"` in `eas.json`, which this PR replaces with the same key plus the `android` block. Nothing else is removed anywhere.
- Parsing the workflow and comparing the `build-and-publish` job object against `origin/main` shows it is **equivalent to main** — signing, ASC key creation, build, artifact upload, submission and the `if: always()` cleanup are all untouched. The lint gate deliberately has no `needs:` edge to the iOS job, so iOS submission is not coupled to it.

**The two-commit split survived**, so the ABI change stays independently revertible:

| Commit | Contents |
|---|---|
| `94d39101` | lint gate — `eas.json` `gradleCommand` exclusions + the `android-release-lint` job |
| `38e55e35` | ABI/NDK — `app.config.js` `buildArchs`, the ABI guard step, the second NDK and `cmake;3.22.1` |

Confirmed by test-reverting `38e55e35` locally: it applies cleanly, the lint gate and the `eas.json` exclusions survive, and #618's cleanup step is still the last iOS step.

**Re-validation after the rebase.** `actionlint` output is byte-identical to the `origin/main` baseline for this file (the single pre-existing `macos-tartelet` custom-label note at `15:42`); YAML parses; assertions confirm both the `android-release-lint` job and the iOS job's final `Remove App Store Connect key` step with `if: always()` are present; `bash -n` clean on all six `run:` blocks; `yarn ts` 13/13, `yarn lint` clean, `yarn deadcode` only pre-existing hints, `yarn cpd` 0 clones; both commit subjects pass commitlint.

Pushed with `--force-with-lease` pinned to the previous head `5f5b77c`. Not merged — the ABI decision still needs the human veto call described in the PR body.

Issue #632, opened 2026-08-08
epic(consolidation): matching correctness, repair, provenance & privacy hardening


vitalyiegorov:
## Why

The consolidation bug-fix series (PRs #602, #603, #611→#607, #612→#606, #613→#608, all merged 2026-07-22…2026-08-02) fixed the reported PrivatBank/Monobank refund and fee defects, but left six open, unparented issues. This epic groups them, records the verified current state, and fixes the execution order.

## Verified state (against `main` @ `ce985a099`, 2026-08-08)

Line-level re-verification of the refund SQL (`packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts`):

- **#604 — STILL VALID.** The `refundCandidateCount = 1` acceptance branch has no expense-side mutual-best requirement. Two refunds whose only candidate is the same expense collapse into one GROUP BY group and `HAVING SUM(refundAmount) <= expenseAmount` **annihilates the whole group** — the legitimate winner is neither consolidated nor reviewable. #612 only changed the HAVING ceiling for the fee bucket; the all-or-nothing semantics is intact. **Adjacent defect found during re-verification:** if the mutual-best winner has `refundCandidateCount > 1` (EXACT_TITLE bucket) while the loser has `= 1`, the loser passes alone and the **wrong refund is auto-consolidated**. Same root cause; must be fixed together.
- **#605 — STILL VALID.** Bucket priority strictly dominates the amount-mismatch sort key in the `refundRank` window ordering, and both consumers hard-filter `refundRank = 1`. Two severity tiers: silent drop (rank-1 bucket is EXACT_TITLE / REJECTED_PAYMENT_* — no review fallback exists for those buckets) and misrouted review (rank-1 is LOCALIZED+mismatch pointing at the wrong expense).
- **#556 — fully specced.** Chain-reclaim repair strategy already designed and validated against a real backup (56/56 chains, fast path, EXPLAIN-indexed).
- **#609 — needs a spec pass** before implementation (provenance, repair tooling, reversibility audits). Should absorb learnings from #604/#605/#556.
- **#610 — scoped audit** of `@Log` hooks in bank-sync/consolidation/contracts; explicitly not a rebuild of closed PR #601.
- **#562 — BLOCKED on PR #561.** No Binance code exists on `main`; every deferred item references code that only lives on `feat/binance-sync`.

## Execution order

| # | Issue | Type | Priority | Constraint |
|---|-------|------|----------|-----------|
[...]

PR #648, opened 2026-08-09
fix(consolidation): stop cross-bucket rank-1 shadowing unique exact-amount refund match

### LINKED ISSUE #605: fix(consolidation): cross-bucket priority can hide a unique localized exact-amount refund match

vitalyiegorov:
## What / Why

`refundRank` (`PARTITION BY refundTxId`) ordered candidates by **bucket priority first**, and both `REFUND_AUTO_CANDIDATES_SQL` and `REFUND_REVIEW_CANDIDATES_SQL` hard-filter `refundRank = 1`. Because `AUTO_REFUND_EXACT_TITLE` imposes **no amount condition**, a same-raw-title expense with a completely different amount seized rank 1 for a refund, hiding that refund's *unique localized exact-amount* match at rank 2 — where neither the auto query nor the review query could reach it.

Fixes #605. Part of epic #632.

## Pre-fix behavior observed (post-#633, measured — not assumed)

Both severity tiers from the issue survived #633's `gated_candidates` / `filled_candidates` rework:

| Scenario | Pre-fix auto | Pre-fix review | Tier |
| --- | --- | --- | --- |
| Refund 340 UAH; shadow `Скасування. UBER TRIP` expense 500 UAH (EXACT_TITLE); true `UBER TRIP` expense 340 UAH (LOCALIZED, exact) | `[]` | one row pointing at the **wrong** (shadow) expense, 340 ≤ 500 passes `HAVING` | **misrouted review** |
| Refund 340 UAH; shadow `Скасування. LIME RIDE` expense 120 UAH (EXACT_TITLE); two `LIME RIDE` expenses 340 UAH (LOCALIZED, ambiguous) | `[]` | `[]` — the rank-1 shadow row is killed by `HAVING SUM(refundAmount) <= expenseAmount` (340 ≤ 120 fails) and rank-2 rows are filtered out | **silent drop** (refund vanishes entirely) |

## Chosen design: option 2 (reorder ranking keys)

`buildLocalizedAmountMismatchOrderSql` (localized-bucket-only, applied *after* bucket priority) is replaced by a bucket-agnostic `buildAmountMismatchOrderSql` placed *before* bucket priority:

```sql
ORDER BY CASE WHEN refundAmount = <REFUND_CEILING_SQL> THEN 0 ELSE 1 END,
         <bucketPriority>, timeDiff
```

Why option 2 over per-bucket ranking (option 1):

- Option 1 (rank per `(refundTxId, confidenceBucket)`, accept in bucket order) **fails the precedence guard**: acceptance would fall through an unacceptable-but-plausible EXACT_TITLE row to a localized-exact-sole row even when the exact-title expense has the *same* exact amount. The real discriminator between "wrong rank-1" and "genuinely better rank-1" is amount-exactness, not bucket membership.
- Option 2 keeps `refundRank = 1` everywhere, so #633's single-winner-per-refund invariant is preserved *by construction* — no new CTE level, no new gate, no risk of a refund being auto-consolidated onto two expenses via two buckets.
- The new key is a strict generalisation of the key it replaces: for `AUTO_REFUND_LOCALIZED_REFUND_TITLE` rows it is identical, and `REJECTED_PAYMENT_PRINCIPAL/FEE` rows are exact by their own bucket conditions (`inc.amount = exp.amount` / `inc.amount = exp.feeAmount`), so only `AUTO_REFUND_EXACT_TITLE` and review rows change relative order — and only against a competing exact-amount candidate.
- Symmetric application to `expenseRank` keeps both windows on the same ordering keys (no divergence between "the refund's best expense" and "the expense's best refund").

## Invariants

| Invariant | Before | After |
| --- | --- | --- |
| At most one auto-eligible row per refund | `refundRank = 1` | unchanged |
| Bucket priority decides between equally exact candidates | yes | yes (bucket key still ahead of `timeDiff`) |
| Amount-exact candidate outranks an inexact one | localized bucket only, below bucket priority | all buckets, above bucket priority |
| Localized-exact-sole auto acceptance (#603) | `localizedExactAmountMatchCount = 1` | unchanged |
| Mutual-best gate / running fill / fee ceiling (#633, #612) | `rejectedBetterRankedRefundCount`, `expenseFilledRefundTotal` | unchanged |
| Refund never vanishes when nothing auto-fires | violated (silent drop) | review falls through to the best exact-amount candidate |

## Test evidence

New file `tests/consolidation-tests/src/scenarios/refund-pair-cross-bucket-rank.test.ts` (written first, confirmed failing on `origin/main`):

- **E** cross-bucket shadow → the unique localized exact-amount match auto-consolidates, review is empty, shadow expense untouched.
- **F** same shadow with ambiguous localized twins → nothing auto-consolidates, the refund still surfaces for review against its nearest exact-amount localized candidate.
- **Guard** equally exact-amount EXACT_TITLE candidate vs localized alternative → EXACT_TITLE stays rank 1, the localized row does not steal the refund, nothing auto-consolidates.

Assertions cover both `refundPairRepository.findCandidates()` / `findReviewCandidates()` and post-`runConsolidation()` database state.

| Suite | Before | After |
| --- | --- | --- |
| `tests/consolidation-tests` | 25 files / 66 tests passing | 26 files / **69 tests passing**, 0 regressions |

Untouched-by-design suites verified green: `refund-pair-competing-refunds`, `refund-pair-localized-exact-amount`, `refund-pair-by-title`, `refund-pair-rejected-payment`(+ statistics), `refund-pair-multiple-refunds`, `refund-pair-manual-review`.

## EXPLAIN QUERY PLAN

Captured for `REFUND_AUTO_CANDIDATES_SQL(null)` and `REFUND_REVIEW_CANDIDATES_SQL` on the migrated test schema, before and after:

- Base-table access unchanged, still all index searches: `transactions_visible_type_operated_idx` for both `income_tx` and `expense_tx`, `transaction_entries_live_transaction_account_amount_idx` for both entry joins, `INTEGER PRIMARY KEY` for the account joins.
- No new full-table scans (`SCAN` line count identical at 20; every `SCAN` is a CTE co-routine, not a table).
- Only delta: the ordering key references the refund ceiling, so SQLite evaluates the `expense_fee_entry` correlated scalar subquery once more per query (9 → 10 instances), each an index search with its existing `LIMIT 1` temp B-tree. No new temp B-tree kinds; `GROUP BY` / `group_concat(ORDER BY)` B-trees unchanged.

## Validation

`yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd` — all clean. No new eslint-disables.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved refund matching by prioritizing candidates with amounts that align with applicable refund limits.
  - Ensured consistent ranking across refund and expense candidates.
  - Unique exact-amount matches can now be consolidated automatically, even when title matches conflict.
  - Ambiguous matches are surfaced for review instead of being consolidated incorrectly.
  - Exact title and amount matches retain priority over localized alternatives.

- **Tests**
  - Added coverage for cross-category refund matching and ranking scenarios.

### DIFF
--- a/packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts
+++ b/packages/consolidation/src/query/repository/sql-factory/refund-ranked-candidate-sql.factory.ts
@@ -191,8 +191,7 @@ const buildBucketPriorityOrderSql = (): string => `
     END
 `;
 
-const buildLocalizedAmountMismatchOrderSql = (): string =>
-    `CASE WHEN confidenceBucket = 'AUTO_REFUND_LOCALIZED_REFUND_TITLE' AND expenseAmount != refundAmount THEN 1 ELSE 0 END`;
+const buildAmountMismatchOrderSql = (): string => `CASE WHEN refundAmount = ${REFUND_CEILING_SQL} THEN 0 ELSE 1 END`;
 
 const buildRankedPairsSql = (): string => `
     ranked_pairs AS (
@@ -201,11 +200,11 @@ const buildRankedPairsSql = (): string => `
             COUNT(*) OVER (PARTITION BY refundTxId) AS refundCandidateCount,
             ROW_NUMBER() OVER (
                 PARTITION BY refundTxId
-                ORDER BY ${buildBucketPriorityOrderSql()}, ${buildLocalizedAmountMismatchOrderSql()}, timeDiff
+                ORDER BY ${buildAmountMismatchOrderSql()}, ${buildBucketPriorityOrderSql()}, timeDiff
             ) AS refundRank,
             ROW_NUMBER() OVER (
                 PARTITION BY expenseTxId
-                ORDER BY ${buildBucketPriorityOrderSql()}, ${buildLocalizedAmountMismatchOrderSql()}, timeDiff, refundTxId
+                ORDER BY ${buildAmountMismatchOrderSql()}, ${buildBucketPriorityOrderSql()}, timeDiff, refundTxId
             ) AS expenseRank
         FROM compatible_pairs
         WHERE confidenceBucket IS NOT NULL
--- a/tests/consolidation-tests/src/scenarios/refund-pair-competing-refunds.test.ts
+++ b/tests/consolidation-tests/src/scenarios/refund-pair-competing-refunds.test.ts
@@ -36,6 +36,24 @@ const NETFLIX_DECOY_EXPENSE_OPERATED_AT = new Date(COMPETING_YEAR, 4, 6, 8, 0, 0
 const NETFLIX_TARGET_EXPENSE_OPERATED_AT = new Date(COMPETING_YEAR, 4, 6, 18, 0, 0);
 const NETFLIX_REFUND_DELAY_SECONDS = 2 * 60 * 60;
 
+const ROZETKA_TITLE = 'ROZETKA';
+const ROZETKA_EXPENSE_AMOUNT_UAH = 500;
+const ROZETKA_EXPENSE_AMOUNT = ROZETKA_EXPENSE_AMOUNT_UAH * PRECISION;
+const ROZETKA_PARTIAL_REFUND_AMOUNT_UAH = 200;
+const ROZETKA_PARTIAL_REFUND_AMOUNT = ROZETKA_PARTIAL_REFUND_AMOUNT_UAH * PRECISION;
+const ROZETKA_EXPENSE_OPERATED_AT = new Date(COMPETING_YEAR, 5, 8, 9, 0, 0);
+const ROZETKA_REFUND_DELAY_SECONDS = 4 * 60 * 60;
+
+const COMFY_TITLE = 'COMFY';
+const COMFY_EXPENSE_AMOUNT_UAH = 900;
+const COMFY_EXPENSE_AMOUNT = COMFY_EXPENSE_AMOUNT_UAH * PRECISION;
+const COMFY_LARGER_REFUND_AMOUNT_UAH = 500;
+const COMFY_LARGER_REFUND_AMOUNT = COMFY_LARGER_REFUND_AMOUNT_UAH * PRECISION;
+const COMFY_SMALLER_REFUND_AMOUNT_UAH = 400;
+const COMFY_SMALLER_REFUND_AMOUNT = COMFY_SMALLER_REFUND_AMOUNT_UAH * PRECISION;
+const COMFY_EXPENSE_OPERATED_AT = new Date(COMPETING_YEAR, 5, 20, 9, 0, 0);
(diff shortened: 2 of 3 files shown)


vitalyiegorov:
Follow-up commit `test(consolidation): pin exact-amount-first expense fill ordering` adds coverage for the symmetric `expenseRank` ordering shift flagged in the PR description.

Added to `tests/consolidation-tests/src/scenarios/refund-pair-competing-refunds.test.ts` (new describe block `expense fill ordering`) — its `seedExactTitleRefunds` fixture already models one expense with several exact-title refunds competing for the ceiling, which is exactly the running-fill shape under test; `refund-pair-cross-bucket-rank.test.ts` is built around a shadow-expense fixture for cross-bucket ranking.

- `fills the expense with its exact-amount refund and reviews the earlier partial that no longer fits` — expense 500 UAH, refunds 200 (earlier partial) + 500 (later exact), SUM 700 > ceiling. The exact-amount refund now takes `expenseRank` 1 and auto-consolidates; the partial overflows to `findReviewCandidates()` rather than being dropped. **Verified failing on the pre-fix ordering** (the earlier partial used to fill first and pushed the exact-amount refund to review).
- `still absorbs both partial refunds when their total fits the expense ceiling` — expense 900 UAH, refunds 500 + 400 = 900, both auto-consolidate, review empty. Passes before and after, guarding multi-refund fill against the ordering change.

Note on the requested non-overflow sibling: a literal "full refund + partial with SUM ≤ ceiling" pair is arithmetically impossible, since an exact-amount refund already consumes the entire ceiling — the guard therefore pins two partials that together fill the expense exactly.

Suite: **26 files / 71 tests passing** (was 69 on the previous commit, 66 on `main`), zero regressions. `yarn format && yarn ts && yarn lint && yarn deadcode && yarn cpd` all clean.

